PR #6, opened 2024-07-05
Streaming write for DuckDB TableProvider

sgrebnov:
PR implements streaming write support for the DuckDB `TableProvider`, which is critical for handling large volumes of data that cannot be stored entirely in memory before writing to a DuckDB file-based `TableProvider`.

Since DuckDB transactions are not `Sync`, they cannot be used in the asynchronous context required for stream reading. To address this, a separate Tokio task has been created for synchronous DuckDB writing. A `tokio` channel with `blocking_recv` is used to propagate `RecordBatches` for writing.

### DIFF
--- a/src/duckdb/write.rs
+++ b/src/duckdb/write.rs
@@ -190,9 +190,9 @@ impl DataSink for DuckDBDataSink {
         }
 
         if notify_commit_transaction.send(()).is_err() {
-            return Err(DataFusionError::Execution(format!(
-                "Unable to send message to commit transaction to duckdb writer."
-            )));
+            return Err(DataFusionError::Execution(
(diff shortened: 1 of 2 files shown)


src/duckdb/write.rs
```diff
@@ -117,42 +119,71 @@ impl DataSink for DuckDBDataSink {
...
+
+                tx.commit()
+                    .context(super::UnableToCommitTransactionSnafu)
+                    .map_err(to_datafusion_error)?;
+
+                Ok(num_rows)
+            });
+
+        while let Some(batch) = data.next().await {
+            let batch = batch.map_err(check_and_mark_retriable_error)?;
```
sgrebnov, line 166: The error should also be passed to duckdb_write_handle task so we know about the error and don't commit/write the data.

PR #21, opened 2024-07-29
Fix schema extraction for MySql tables

sgrebnov:
PRs fixes schema extraction for MySql empty tables.

Previously, when a table didn't have any records during the connection, an empty (invalid) schema was returned. This PR switches to use `SHOW COLUMNS` (table describe) query to extract the schema. This is now default approach to extract table schema.

Verified that TPC-H tests pass.

~~It is proposed to not replace the original schema extraction mechanism immediately, allowing us to test and evaluate the new approach before adopting it as the primary method in the future.~~

### DIFF
--- a/src/sql/db_connection_pool/dbconnection/mysqlconn.rs
+++ b/src/sql/db_connection_pool/dbconnection/mysqlconn.rs
@@ -1,7 +1,8 @@
 use std::{any::Any, sync::Arc};
 
+use crate::sql::arrow_sql_gen::mysql::map_column_to_data_type;
 use crate::sql::arrow_sql_gen::{self, mysql::rows_to_arrow};
-use arrow::datatypes::{Schema, SchemaRef};
+use arrow::datatypes::{DataType, Field, Schema, SchemaRef};
 use async_stream::stream;
 use datafusion::error::DataFusionError;
 use datafusion::execution::SendableRecordBatchStream;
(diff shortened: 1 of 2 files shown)


src/sql/db_connection_pool/dbconnection/mysqlconn.rs
```diff
@@ -72,11 +83,63 @@ impl<'a> AsyncDbConnection<Conn, &'a (dyn ToValue + Sync)> for MySQLConnection {
...
+            let rec = rows_to_arrow(&rows)
+                .boxed()
+                .context(super::UnableToGetSchemaSnafu)?;
+            return Ok(rec.schema());
+        }
+
+        // we don't use SELECT COLUMN_NAME, DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = '{}' AND TABLE_SCHEMA = '{}'
+        // as table_reference don't always have schema specified so we need to extract schema/db name from connection properties
+        // to ensure we are querying information for correct table
+        let metadata_query = format!("SHOW COLUMNS FROM {}", table_reference.to_quoted_string());
```
phillipleblanc, line 96: Let's refactor this into a separate method, similar to `rows_to_arrow` we can call. Maybe `metadata_rows_to_schema`?

sgrebnov: Updated - moved to `columns_meta_to_schema` fn

lukekim (approved):

phillipleblanc (approved):

PR #24, opened 2024-08-01
feat: Stream SQLite writes

peasee:
## 🗣 Description

* Converts the SQLite writer to stream writes instead of collecting all of the data to write in memory and then writing it

Through some manual testing, this improves both write performance (by 20-40%) and memory usage (by 80+%) during write operations with 60 million row tables.


phillipleblanc (approved):

src/sqlite/write.rs
```diff
@@ -156,9 +158,9 @@ impl DataSink for SqliteDataSink {
                     sqlite.delete_all_table_data(&transaction)?;
                 }
 
-                for batch in data_batches {
-                    if batch.num_rows() > 0 {
-                        sqlite.insert_batch(&transaction, batch, on_conflict.as_ref())?;
+                while let Some(data_batch) = batch_rx.blocking_recv() {
```
sgrebnov, line 161: @peasee - can we perform a read query while streaming results?

peasee: You can perform read queries for tables except the ones still accelerating. If you try querying a table that's accelerating, the query will hang until the acceleration finishes before returning values.

This is the existing behavior on `main`.

sgrebnov:
This is different as we previously collected all data first and was writing only after that thus less locking time

Thanks,
Sergey
________________________________

Sevenannn (approved):

PR #28, opened 2024-08-02
Support mysql BINARY, VARBINARY and Postgres Type::BYTEA

Sevenannn:
Support mysql BINARY, VARBINARY and Postgres Type::BYTEA for the mysql table provider and postgres table provider
* mysql `BINARY` & `CHAR` are both of type `MYSQL_TYPE_STRING` in mysql-async library. Add additional binary flag data to determine mysql binary / char and cast them to arrow binary / largestring accordingly

### DIFF
--- a/src/sql/arrow_sql_gen/mysql.rs
+++ b/src/sql/arrow_sql_gen/mysql.rs
@@ -461,7 +461,7 @@ pub fn map_column_to_data_type(
         | ColumnType::MYSQL_TYPE_LONG_BLOB => Some(DataType::LargeUtf8),
         ColumnType::MYSQL_TYPE_STRING
         | ColumnType::MYSQL_TYPE_VAR_STRING => {
-            if column_flag {
+            if column_is_binary {
                 Some(DataType::Binary)
             } else {
                 Some(DataType::LargeUtf8)
--- a/src/sql/arrow_sql_gen/mysql.rs
+++ b/src/sql/arrow_sql_gen/mysql.rs
@@ -1,10 +1,10 @@
 use crate::sql::arrow_sql_gen::arrow::map_data_type_to_array_builder_optional;
 use arrow::{
     array::{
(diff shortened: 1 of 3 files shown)


src/sql/arrow_sql_gen/mysql.rs
```diff
@@ -275,6 +273,10 @@ pub fn rows_to_arrow(rows: &[Row]) -> Result<RecordBatch> {
                         i
                     );
                 }
+                column_type @ (ColumnType::MYSQL_TYPE_STRING
```
sgrebnov, line 276: 1. Is this for acceleration? 
2. Could you please verify that query returns the same result as direct query
3. As we are changing to MYSQL_TYPE_STRING to be binary I propose to change  MYSQL_TYPE_STRING->MYSQL_TYPE_VARCHAR for `char`
_ if data_type.starts_with("char") => ColumnType::MYSQL_TYPE_STRING,

sgrebnov: Example query
image

sgrebnov (approved):

src/sql/arrow_sql_gen/mysql.rs
Sevenannn: This is for acceleration, the verified data can be seen from spice oss ticket

Sevenannn: Checked with the new fix, it doesn't broke query in spice repl, see the query result in: https://github.com/spiceai/spiceai/issues/1871

Sevenannn: Added new code to detect whether the column is binary (BINARY, VARBINARY) or non binary (CHAR VARCHAR) . The change don't have any degrading behavior tested on tpch queries

sgrebnov (approved):
@Sevenannn  -  great catch on BINARY_FLAG 👍

phillipleblanc (approved):

PR #45, opened 2024-08-16
test: Add SQLite-Arrow roundtrip integration test

peasee:
## 🗣 Description

* Adds round-trip integration test for testing Arrow data types in SQLite.

By default ignores the following tests as they fail: `struct`, `decimal`, `interval`, `duration`, `list`.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -59,4 +59,4 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
 
       - name: Run integration test
-        run: cargo test --test integration --no-default-features --features postgres -- --nocapture
+        run: make test-integration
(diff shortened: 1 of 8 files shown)


Sevenannn:
Changes look good to me. Some minor suggestions - maybe we can wrap all round trip test inside one function call so that we don't need init tracing for so many times. Also we can abstract the database creation part out so that we don't need to create a new sqlite database for each type of table.

phillipleblanc (approved):

peasee:
> Changes look good to me. Some minor suggestions - maybe we can wrap all round trip test inside one function call so that we don't need init tracing for so many times. Also we can abstract the database creation part out so that we don't need to create a new sqlite database for each type of table.

For SQLite, we're only loading an in-memory database and it validates that we have a fresh DB for each test. The tests run nearly immediately, so I don't think test performance is a concern.

phillipleblanc (approved):

phillipleblanc (approved):

PR #52, opened 2024-08-19
fix: Don't attach empty SQLite attachments

peasee:
## 🗣 Description

* Fixes SQLite database attachments to avoid attempting to attach when an empty vec of attachments is supplied.
* Updates tests to use randomised DB filenames to make new tests easier to add without clashing filenames.

### DIFF
--- a/src/sql/db_connection_pool/sqlitepool.rs
+++ b/src/sql/db_connection_pool/sqlitepool.rs
@@ -215,81 +215,125 @@ impl DbConnectionPool<Connection, &'static (dyn ToSql + Sync)> for SqliteConnect
 mod tests {
     use super::*;
     use crate::sql::db_connection_pool::Mode;
+    use rand::Rng;
+    use rstest::{fixture, rstest};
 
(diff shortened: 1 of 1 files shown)


src/sql/db_connection_pool/sqlitepool.rs
```diff
@@ -46,26 +46,30 @@ impl SqliteConnectionPoolFactory {
...
+                        attach_databases.push(Arc::clone(&self.path));
+                    }
 
-                attach_databases.sort();
+                    attach_databases.sort();
 
-                JoinPushDown::AllowedFor(attach_databases.join(";")) // push down is allowed cross-database when they're attached together
-                                                                     // hash the list of databases to generate the comparison for push down
+                    JoinPushDown::AllowedFor(attach_databases.join(";")) // push down is allowed cross-database when they're attached together
+                                                                         // hash the list of databases to generate the comparison for push down
```
phillipleblanc, line 71:
```suggestion
```

phillipleblanc (approved):

PR #54, opened 2024-08-20
feat: support postgres geometry & BYTEA_ARRAY

xhwhis:


src/sql/arrow_sql_gen/postgres.rs
```diff
@@ -721,6 +775,15 @@ fn map_column_type_to_data_type(column_type: &Type) -> Option<DataType> {
             DataType::Boolean,
             true,
         )))),
+        Type::BYTEA_ARRAY => Some(DataType::List(Arc::new(Field::new(
```
phillipleblanc, line 778: Would you be able to add a test for this so we can ensure it doesn't break in the future? Adding a new test method to https://github.com/datafusion-contrib/datafusion-table-providers/blob/main/tests/postgres/mod.rs and following the existing pattern for how `test_arrow_postgres_roundtrip` works should be fairly straightforward.

phillipleblanc (commented):
Thanks for the contribution! Once this is updated with a test, I'll merge it.

src/sql/arrow_sql_gen/postgres.rs
xhwhis: I have added tests.

phillipleblanc (approved):

phillipleblanc:
@xhwhis Merged! Thanks again 🙏

PR #59, opened 2024-08-21
Preserve timezone information in arrow type - pg type conversion

Sevenannn:
* Preserve timezone information when converting arrow timestamp(timeunit, timezone) type to sql rows
* Correctly convert Postgres Timestamptz type back to arrow type with time zone information, this fixes the incorrect conversion from pg row -> arrow caused by neglecting timezone information
* Refactor the roundtrip integration test to values check using try_cast_to

### DIFF
--- a/src/sql/arrow_sql_gen/postgres.rs
+++ b/src/sql/arrow_sql_gen/postgres.rs
@@ -18,7 +18,7 @@ use bigdecimal::num_bigint::Sign;
 use bigdecimal::BigDecimal;
 use bigdecimal::ToPrimitive;
 use byteorder::{BigEndian, ReadBytesExt};
-use chrono::Timelike;
+use chrono::{DateTime, Offset, Timelike, Utc};
 use composite::CompositeType;
 use geo_types::geometry::Point;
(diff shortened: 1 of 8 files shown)


src/sql/arrow_sql_gen/postgres.rs
```diff
@@ -484,6 +484,55 @@ pub fn rows_to_arrow(rows: &[Row]) -> Result<RecordBatch> {
                         None => builder.append_null(),
                     }
                 }
+                Type::TIMESTAMPTZ => {
```
phillipleblanc, line 487: Let's add an integration test for this

Sevenannn: Postgres timestamptz -> arrow is covered as part of the round trip integration test as shown in the screenshot. Shall we create a new one-way test just for pg timestamptz row -> arrow, or use the existing integration test.
image

Sevenannn: Value checking addressed in the new commit

phillipleblanc (approved):

PR #67, opened 2024-08-22
Add support for writing Lists/Arrays to SQLite via JSON conversion

sgrebnov:
SQLite does not support Arrays natively so PRs adds fallback on using JSON column type and serialization to JSON string for Arrow List types. With this change previously unsupported types will be stored and retrieved as Ut8 string when queried directly or automatically converted back to Arrow List type when using casting layer:  [Add List types parsing support to schema cast](https://github.com/spiceai/datafusion-federation/pull/13)

List, LargeList, FixedSizeList are supported

### DIFF
--- a/src/sql/arrow_sql_gen/statement.rs
+++ b/src/sql/arrow_sql_gen/statement.rs
@@ -1,5 +1,9 @@
 use arrow::{
-    array::{array, Array, RecordBatch},
+    array::{
+        array, Array, BooleanArray, Float32Array, Float64Array, Int16Array, Int32Array, Int64Array,
+        Int8Array, LargeStringArray, RecordBatch, StringArray, UInt16Array, UInt32Array,
+        UInt64Array, UInt8Array,
+    },
     datatypes::{DataType, Field, IntervalUnit, SchemaRef, TimeUnit},
     util::display::array_value_to_string,
 };
@@ -11,7 +15,7 @@ use sea_query::{
     QueryBuilder, SimpleExpr, SqliteQueryBuilder, StringLen, Table,
 };
 use snafu::Snafu;
-use std::sync::Arc;
+use std::{any::Any, sync::Arc};
 use time::{OffsetDateTime, PrimitiveDateTime};
 
 #[derive(Debug, Snafu)]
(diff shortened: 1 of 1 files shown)


src/sql/arrow_sql_gen/statement.rs
```diff
@@ -894,6 +942,63 @@ fn insert_timestamp_into_row_values(
...
+    }};
+}
+
+
+fn insert_list_into_row_values_json(
+    list_array: Arc<dyn Array>,
+    list_type: &Arc<Field>,
+    row_values: &mut Vec<SimpleExpr>,
+) -> Result<()> {
+    // arrow_json based serialization
```
sgrebnov, line 968: Note: can potentially use `arrow_json::ArrayWriter`, but it produces a non-flat array, which is probably good for some complex types serialization but not for Arrays/Lists IMO

```
"[{"item":1},{"item":2},{"item":3}]"
```

src/sql/arrow_sql_gen/statement.rs
```diff
@@ -166,6 +176,20 @@ pub struct InsertBuilder {
     record_batches: Vec<RecordBatch>,
 }
 
+pub fn use_json_insert_for_type<T: QueryBuilder + 'static>(
```
sgrebnov, line 179: static lifetime marker is required for downcasting to detect SQLite engine

src/sql/arrow_sql_gen/statement.rs
```diff
@@ -88,6 +92,12 @@ impl CreateTableBuilder {
     #[must_use]
     pub fn build_sqlite(self) -> String {
         self.build(SqliteQueryBuilder, &|f: &Arc<Field>| -> ColumnType {
+            // Sqlite does not natively support Arrays, Structs, etc
+            // so we use JSON column type for List, FixedSizeList, LargeList, Struct, etc
+            if f.data_type().is_nested() {
```
Sevenannn, line 97: Arrow map and union types are also nested, we can make a note inside the comments here, so when later on union / map is supported for sqlite, it's easier to locate this relevant changes

Sevenannn (approved):

phillipleblanc (approved):

PR #70, opened 2024-08-25
Add Structs support for SQLite

sgrebnov:
SQLite does not support  Structs natively so PRs adds fallback on using JSON column type and serialization to JSON string for Arrow struct type. 

With this change previously unsupported Arrow Struct type will be stored and retrieved as Utf8 string when queried directly or automatically converted to Arrow Struct type when using casting layer:  https://github.com/spiceai/datafusion-federation/pull/16

Integration tests for struct will be enabled after casting PR is merged and reference is updated: https://github.com/spiceai/datafusion-federation/pull/16

PR also switches from own `try_cast_to` implementation to`try_cast_to` exposed by datafusion-federation

### DIFF
--- a/src/sql/arrow_sql_gen/statement.rs
+++ b/src/sql/arrow_sql_gen/statement.rs
@@ -1,10 +1,10 @@
 use arrow::{
     array::{
-        array, Array, BooleanArray, Float32Array, Float64Array, Int16Array, Int32Array, Int64Array,
-        Int8Array, LargeStringArray, RecordBatch, StringArray, UInt16Array, UInt32Array,
-        UInt64Array, UInt8Array,
+        array, Array, ArrayRef, BooleanArray, Float32Array, Float64Array, Int16Array, Int32Array,
+        Int64Array, Int8Array, LargeStringArray, RecordBatch, StringArray, StructArray,
+        UInt16Array, UInt32Array, UInt64Array, UInt8Array,
     },
-    datatypes::{DataType, Field, IntervalUnit, SchemaRef, TimeUnit},
+    datatypes::{DataType, Field, Fields, IntervalUnit, Schema, SchemaRef, TimeUnit},
     util::display::array_value_to_string,
 };
 use bigdecimal_0_3_0::BigDecimal;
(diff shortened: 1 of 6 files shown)


tests/arrow_record_batch_gen/mod.rs
```diff
@@ -277,30 +276,63 @@ pub(crate) fn get_arrow_date_record_batch() -> (RecordBatch, SchemaRef) {
...
+        .field_builder::<BooleanBuilder>(0)
+        .expect("should return field builder")
+        .append_value(false);
+    struct_builder
+        .field_builder::<Int32Builder>(1)
+        .expect("should return field builder")
+        .append_value(30);
+    struct_builder.append(true);
+
+    // NULL struct item is temporary disabled as not properly supported by duckdb and postgres
```
sgrebnov, line 312: Updated get_arrow_struct_record_batch in order to be able to include test for None struct value. Tested for sqlite (works) but temporary disabled as duckdb and postgres don't handle it correctly. To be investigated.

src/sql/arrow_sql_gen/statement.rs
```diff
@@ -1312,6 +1322,41 @@ fn insert_list_into_row_values_json(
...
+) -> Result<()> {
+    // The length of each column in a StructArray is the same as the length of the StructArray itself.
+    // The presence of null values does not change the length of the columns (affects the validity bitmap only).
+    // Similar to Recordbatch slice: https://github.com/apache/arrow-rs/blob/855666d9e9283c1ef11648762fe92c7c188b68f1/arrow-array/src/record_batch.rs#L375
+    let single_row_columns: Vec<ArrayRef> = (0..array.num_columns())
+        .map(|i| array.column(i).slice(row_index, 1))
+        .collect();
+
+    let batch =
+        RecordBatch::try_new(Arc::new(Schema::new(fields.clone())), single_row_columns).unwrap();
```
phillipleblanc, line 1339: Let's avoid using `.unwrap()`

sgrebnov: Updated, thank you

phillipleblanc (approved):

PR #77, opened 2024-08-27
Correctly covert timestamp scalarvalue to postgres sql statement

Sevenannn:
* Always use timestamp_nano for timestamp conversions to avoid precision loss
* Add engine information to Postgres Table Provider to have custom Expr => sql
- part of: https://github.com/spiceai/spiceai/pull/2392

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ Cargo.lock
 .vscode/settings.json
 
 .DS_Store
+
+.idea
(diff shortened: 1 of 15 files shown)


phillipleblanc (commented):
Let's add a test for these changes, but otherwise LGTM.

sgrebnov (approved):

sgrebnov (approved):

PR #79, opened 2024-08-30
Fix MySQL connector to respect the source column's decimal precision

sgrebnov:
Fix MySQL connector to respect the source column's decimal precision. The test arrow_mysql_one_way is updated to verify the actual precision.

The query logic is now aware of the target expected (projected) schema, as this is required to get the correct precision for the decimal MySQL column during a query since precision information is not available as part of the MySQL response column type.

### DIFF
--- a/src/sql/arrow_sql_gen/mysql.rs
+++ b/src/sql/arrow_sql_gen/mysql.rs
@@ -101,12 +101,20 @@ pub fn rows_to_arrow(rows: &[Row]) -> Result<RecordBatch> {
             let column_name = column.name_str();
             let column_type = column.column_type();
             let column_is_binary = column.flags().contains(ColumnFlags::BINARY_FLAG);
-            let column_decimal_precision: i8 = column.decimals().try_into().unwrap_or_default();
+
+            let (decimal_precision, decimal_scale) = match column_type {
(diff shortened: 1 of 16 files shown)


src/sql/arrow_sql_gen/mysql.rs
```diff
@@ -475,6 +490,16 @@ fn to_decimal_128(decimal: &BigDecimal, scale: i64) -> Option<i128> {
     (decimal * 10i128.pow(scale.try_into().unwrap_or_default())).to_i128()
 }
 
+fn get_decimal_column_precision(column_name: &str, projected_schema: &SchemaRef) -> Option<u8> {
+
```
phillipleblanc, line 494:
```suggestion
```

src/sql/arrow_sql_gen/mysql.rs
```diff
@@ -475,6 +490,16 @@ fn to_decimal_128(decimal: &BigDecimal, scale: i64) -> Option<i128> {
...
+    let field = projected_schema.field_with_name(column_name).ok()?;
+    match field.data_type() {
+        DataType::Decimal128(precision, _) => Some(*precision),
+        _ => None,
+
```
phillipleblanc, line 499:
```suggestion
```

phillipleblanc (approved):

PR #82, opened 2024-09-01
refactor: Create an SQLiteTable for SQLite SQLExec

peasee:
## 🗣 Description

* Creates a struct, `SQLiteTable`, for creating an SQLite table provider. This follows the existing behavior for `DuckDBTable`, instead of using the generic base `SQLTable`.

### DIFF
--- a/src/sqlite/sql_table.rs
+++ b/src/sqlite/sql_table.rs
@@ -8,7 +8,7 @@ use std::fmt::Display;
 use std::{any::Any, fmt, sync::Arc};
 
 use crate::sql::sql_provider_datafusion::{
-    expr, get_stream, to_execution_error, Result as SqlResult, SqlExec, SqlTable,
+    get_stream, to_execution_error, Result as SqlResult, SqlExec, SqlTable,
 };
 use datafusion::{
     arrow::datatypes::SchemaRef,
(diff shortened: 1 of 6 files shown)


src/sqlite/sql_table.rs
```diff
@@ -0,0 +1,188 @@
...
+    sql::TableReference,
+};
+
+pub struct SQLiteTable<T: 'static, P: 'static> {
+    pub(crate) base_table: SqlTable<T, P>,
+}
+
+impl<T, P> SQLiteTable<T, P> {
+    pub fn new_with_schema(
+        name: &'static str,
```
phillipleblanc, line 32:
```suggestion
```

This will always be `sqlite` right?

src/sqlite/sql_table.rs
```diff
@@ -0,0 +1,188 @@
...
+        pool: &Arc<dyn DbConnectionPool<T, P> + Send + Sync>,
+        schema: impl Into<SchemaRef>,
+        table_reference: impl Into<TableReference>,
+        engine: Option<expr::Engine>,
```
phillipleblanc, line 36: This is also always the same engine right?

phillipleblanc (approved):

PR #88, opened 2024-09-06
SQLite: Validate expected indexes when attaching local datasets

sgrebnov:
PRs adds verification of expected indexes when attaching local datasets to SQLite. This change ensures that the schema of the local copy of a table matches the expected configuration.

Note: created composite indexes can have complex names, produced index name information is easy to troubleshoot IMO so converting index name back to original configuration for tracing has not been added (NOTE: it is  also not always possible or trivial, for example I have the following index `i_taxi_trips_fare_amount_passenger_count_total_amount_tpep_dropoff_datetime_tpep_pickup_datetime_trip_distance`) generated from test configuration `'(tpep_pickup_datetime, tpep_dropoff_datetime, trip_distance, fare_amount, passenger_count, total_amount)': unique`

Sample output:

```shell
2024-09-06T21:46:12.752546Z  WARN datafusion_table_providers::sqlite: Schema mismatch detected for table 'taxi_trips'. The following expected indexes are missing: ["i_taxi_trips_passenger_count"].
2024-09-06T21:46:12.752596Z  WARN datafusion_table_providers::sqlite: Schema mismatch detected for table 'taxi_trips'. The table contains unexpected indexes not defined in the configuration: ["i_taxi_trips_VendorID"].
2024-09-06T21:46:12.752620Z  WARN datafusion_table_providers::sqlite: The schema of the local copy of table 'taxi_trips' does not match the expected configuration. To correct this, you can drop the existing local copy at '/Users/sg/spice/spiceai/.spice/data/taxi_trips_sqlite.db'. A new table will be automatically created with the correct schema upon the first access.
2024-09-06T21:46:12.753283Z  INFO runtime: Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), acceleration (sqlite:file), results cache enabled.
```

In progress: unit test for `get_indexes`

Note: similar improvement must be done for `primary_key`

### DIFF
--- a/src/sqlite.rs
+++ b/src/sqlite.rs
@@ -1,5 +1,5 @@
 use crate::sql::arrow_sql_gen::statement::{CreateTableBuilder, IndexBuilder, InsertBuilder};
-use crate::sql::db_connection_pool::dbconnection::{self, get_schema};
+use crate::sql::db_connection_pool::dbconnection::{self, get_schema, AsyncDbConnection};
 use crate::sql::db_connection_pool::sqlitepool::SqliteConnectionPoolFactory;
 use crate::sql::db_connection_pool::{
     self,
@@ -8,6 +8,7 @@ use crate::sql::db_connection_pool::{
     DbConnectionPool, Mode,
 };
 use crate::sql::sql_provider_datafusion;
+use arrow::array::StringArray;
 use arrow::{array::RecordBatch, datatypes::SchemaRef};
 use async_trait::async_trait;
 use datafusion::catalog::Session;
@@ -19,9 +20,11 @@ use datafusion::{
     logical_expr::CreateExternalTable,
     sql::TableReference,
 };
+use futures::TryStreamExt;
 use rusqlite::{ToSql, Transaction};
 use snafu::prelude::*;
 use sql_table::SQLiteTable;
+use std::collections::HashSet;
 use std::{collections::HashMap, sync::Arc};
 use tokio_rusqlite::Connection;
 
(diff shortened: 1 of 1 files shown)


src/sqlite.rs
```diff
@@ -442,4 +450,77 @@ impl Sqlite {
...
+
+        let missing_in_actual = expected_indexes_str_map
+            .difference(&actual_indexes_str_map)
+            .collect::<Vec<_>>();
+        let extra_in_actual = actual_indexes_str_map
+            .difference(&expected_indexes_str_map)
+            .collect::<Vec<_>>();
+
+        if !missing_in_actual.is_empty() {
+            tracing::warn!(
```
sgrebnov, line 510: Not sure we should warn here, we might just want to return back detected issues.

src/sqlite.rs
```diff
@@ -442,4 +450,77 @@ impl Sqlite {
 
         Ok(())
     }
+
+    async fn get_indexes(
```
sgrebnov, line 454: Note: We collect all names without information on whether an index is unique. This should be sufficient for most cases without adding additional complexity.

src/sqlite.rs
Sevenannn: The warning should probably just explicitly state index instead of `schema mismatch`. Index is not part of the table schema in SQLite, instead they are separate database object

src/sqlite.rs
```diff
@@ -246,6 +249,11 @@ impl TableProviderFactory for SqliteTableProviderFactory {
                 .await
                 .context(UnableToCreateTableSnafu)
                 .map_err(to_datafusion_error)?;
+        } else if !sqlite.verify_indexes_match(sqlite_conn, &indexes).await? {
+            tracing::warn!(
+                "The schema of the local copy of table '{name}' does not match the expected configuration. To correct this, you can drop the existing local copy at '{db_path}'. A new table will be automatically created with the correct schema upon the first access.",
```
Sevenannn, line 254: Same here on the `schema`, should only be `index`

src/sqlite.rs
sgrebnov: Updated - thank you

src/sqlite.rs
sgrebnov: Updated - thank you

phillipleblanc (approved):

PR #91, opened 2024-09-10
fix: Don't silence blocking task errors

peasee:
## 🗣 Description

* Ensures that errors from within blocking tasks don't get silenced, as the return type from `JoinHandle` is `Result<T, JoinError>`. If the inner blocking tasks return a `Result<>` themselves, we need to ensure we match for `Ok(Err(e)` to get the inner error, or `Err(e)` to get the `JoinError`.

### DIFF
--- a/src/duckdb/write.rs
+++ b/src/duckdb/write.rs
@@ -184,11 +184,20 @@ impl DataSink for DuckDBDataSink {
             .context(super::ConstraintViolationSnafu)
             .map_err(to_datafusion_error)?;
 
-            batch_tx.send(batch).await.map_err(|e| {
-                DataFusionError::Execution(format!(
-                    "Unable to send RecordBatch to duckdb writer: {e}"
-                ))
-            })?;
+            if let Err(e) = batch_tx.send(batch).await {
+                match duckdb_write_handle.await {
+                    Err(e) => {
+                        return Err(DataFusionError::Execution(format!(
+                            "Error writing to DuckDB: {e}"
+                        )));
+                    }
+                    _ => {
+                        return Err(DataFusionError::Execution(format!(
+                            "Unable to send RecordBatch to duckdb writer: {e}"
+                        )))
+                    }
+                };
+            }
         }
 
         if notify_commit_transaction.send(()).is_err() {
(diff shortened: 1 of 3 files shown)


src/duckdb/write.rs
```diff
@@ -184,11 +184,25 @@ impl DataSink for DuckDBDataSink {
...
+                        )));
+                    }
+                    Ok(Err(e)) => {
+                        return Err(DataFusionError::Execution(format!(
+                            "Error writing to DuckDB: {e}"
+                        )));
+                    }
+                    _ => {
+                        return Err(DataFusionError::Execution(format!(
+                            "Unable to send RecordBatch to duckdb writer: {e}"
```
sgrebnov, line 201:
```suggestion
                            "Unable to send RecordBatch to DuckDB writer: {e}"
```

sgrebnov (approved):

src/duckdb/write.rs
```diff
@@ -184,11 +184,25 @@ impl DataSink for DuckDBDataSink {
...
+                    Err(e) => {
+                        return Err(DataFusionError::Execution(format!(
+                            "Error writing to DuckDB: {e}"
+                        )));
+                    }
+                    Ok(Err(e)) => {
+                        return Err(DataFusionError::Execution(format!(
+                            "Error writing to DuckDB: {e}"
+                        )));
+                    }
```
phillipleblanc, line 198: This overloading of `e` is a bit confusing, can you rename them?

src/sql/db_connection_pool/dbconnection/duckdbconn.rs
```diff
@@ -151,10 +151,18 @@ impl SyncDbConnection<r2d2::PooledConnection<DuckdbConnectionManager>, DuckDBPar
...
+            match join_handle.await {
+                Ok(Err(e)) => {
+                    yield Err(DataFusionError::Execution(format!(
+                        "Failed to execute DuckDB query: {e}"
+                    )))
+                },
+                Err(e) => {
+                    yield Err(DataFusionError::Execution(format!(
+                        "Failed to execute DuckDB query: {e}"
+                    )))
```
phillipleblanc, line 163: What is the difference between these cases?

peasee: `Err(e)` is the `JoinError` from the `JoinHandle`, which would usually trigger if e.g. tokio couldn't spawn the task or something. `Ok(Err(e))` is where tokio has successfully spawned the task, but there's an error from within the task.

phillipleblanc (approved):

PR #98, opened 2024-09-14
Fix SQLite `Invalid column type Real` error

sgrebnov:
As detailed at [Bug: SQLite acceleration fails with Invalid column type Real TPCH Q10 and Q19](https://github.com/spiceai/spiceai/issues/2638) the SQLite Table Provider was failing with the following error when attempting to read a Float value as an Integer. This issue occurs when a query returns the first row for a Float column with a value without a fractional component, which SQLite represents as an Integer.

>Execution error: Unable to query arrow: ConnectionError Other("Failed to convert query result to Arrow: Failed to extract row value: Invalid column type Real at index: 3, name: l_discount")

The PR updated the schema inference to use the projected schema to accurately determine if the column is actually a Float.

Note: The error has a stable repro but occurs only in in-memory mode.

### DIFF
--- a/src/sql/db_connection_pool/dbconnection/sqliteconn.rs
+++ b/src/sql/db_connection_pool/dbconnection/sqliteconn.rs
@@ -63,7 +63,7 @@ impl AsyncDbConnection<Connection, &'static (dyn ToSql + Sync)> for SqliteConnec
                 let mut stmt = conn.prepare(&format!("SELECT * FROM {table_reference} LIMIT 1"))?;
                 let column_count = stmt.column_count();
                 let rows = stmt.query([])?;
-                let rec = rows_to_arrow(rows, column_count)
+                let rec = rows_to_arrow(rows, column_count, None)
                     .context(ConversionSnafu)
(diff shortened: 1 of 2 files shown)


src/sql/db_connection_pool/dbconnection/sqliteconn.rs
```diff
@@ -63,7 +63,7 @@ impl AsyncDbConnection<Connection, &'static (dyn ToSql + Sync)> for SqliteConnec
                 let mut stmt = conn.prepare(&format!("SELECT * FROM {table_reference} LIMIT 1"))?;
                 let column_count = stmt.column_count();
                 let rows = stmt.query([])?;
-                let rec = rows_to_arrow(rows, column_count)
+                let rec = rows_to_arrow(rows, column_count, None)
```
sgrebnov, line 66: For acceleration (write) mode this is not used so there is no such issue. For SQLite data connector mode this should be improved to rely on native schema inference instead of using first row

phillipleblanc (approved):

Issue #18, opened 2024-07-25
Use DataFusion unparser instead of sql_provider_datafusion/expr.rs


phillipleblanc:
There is basic support for converting DataFusion `expr` to SQL as part of https://github.com/datafusion-contrib/datafusion-table-providers/blob/main/src/sql/sql_provider_datafusion/expr.rs

We should just be using the [DataFusion unparser](https://github.com/apache/datafusion/blob/main/datafusion/sql/src/unparser/expr.rs#L105) directly and remove this crate's version.

hozan23:
@phillipleblanc I think we can close this issue ref: #99, #102

PR #99, opened 2024-09-14
use datafusion::Unparser for converting datafusion Expr to SQL

hozan23:
Hello, 

This PR addresses issue #18.

It removes the basic unparser that was used to convert DataFusion's `Expr` into SQL and instead uses the implementation from [DataFusion's unparser](https://github.com/apache/datafusion/blob/main/datafusion/sql/src/unparser/expr.rs#L91).

Additionally, this PR modifies the API for the `SqlTable` and `SqlExec` structs to accept a SQL `Dialect`, which is used for parsing.


src/sql/sql_provider_datafusion/mod.rs
```diff
@@ -45,7 +46,27 @@ pub enum Error {
...
+}
+
+#[derive(Clone, Copy, Debug)]
+pub enum Engine {
+    Spark,
+    SQLite,
+    DuckDB,
+    ODBC,
+    Postgres,
+}
```
phillipleblanc, line 59: I don't think this `Engine` concept is still needed now that we've moved to the DataFusion Unparser - it served basically the same purpose as the Dialect.

src/sql/sql_provider_datafusion/mod.rs
```diff
@@ -76,38 +98,37 @@ impl<T, P> SqlTable<T, P> {
...
-            dialect: None,
-        }
-    }
-
-    #[must_use]
-    pub fn with_dialect(self, dialect: Arc<dyn Dialect + Send + Sync>) -> Self {
-        Self {
-            dialect: Some(dialect),
-            ..self
+            dialect,
```
phillipleblanc, line 131: I would prefer to keep the `with_dialect` function. This makes it cleaner to override the default dialect if required without cluttering the main constructor. The main constructor can now lose two arguments (for engine and dialect).

src/sql/sql_provider_datafusion/mod.rs
```diff
@@ -276,24 +299,35 @@ impl<T, P> SqlExec<T, P> {
...
+    fn table_name_escaped(&self) -> String {
+        self.table_reference.to_quoted_string()
+    }
+
+    fn column_name_escaped(&self, column_name: &str) -> String {
+        match self.engine {
+            Some(Engine::ODBC) => column_name.to_string(),
+            _ => {
+                format!("\"{}\"", column_name)
+            }
```
phillipleblanc, line 328: This is probably the only reason we have the Engine now right? I'm pretty sure the Dialect has a method to determine the column escaping character that we could use.

phillipleblanc (approved):
Thanks for the PR!

hozan23:
Hey @phillipleblanc,

I was also considering the need of Engine, but I'm thinking of wrapping the few Dialects offered by DataFusion with the current Engine enum we have. This way, we can still add some custom logic to these engines and not be restricted by the types offered by DataFusion's dialects. By doing this, we would no longer need to use Dialect directly, and we could remove it from `SqlTable` and `SqlExec`, no?

phillipleblanc:
I pushed up a few commits to implement the Engine -> Dialect mapping and removed passing both Engine & Dialect.

phillipleblanc:
Thanks @hozan23!

hozan23:
Thanks for merging it @phillipleblanc

PR #100, opened 2024-09-14
Postgres enum support

Sevenannn:
This PR enables the support for Postgres enum types
* Create EnumValueFromSql to parse Postgres enum types
* Parse Postgres enum types into arrow StringDictionary
* Add a postgres row -> arrow test to validate the cast

### DIFF
--- a/src/sql/arrow_sql_gen/postgres.rs
+++ b/src/sql/arrow_sql_gen/postgres.rs
@@ -1047,7 +1047,7 @@ impl<'a> FromSql<'a> for EnumValueFromSql {
     fn from_sql(
         _ty: &Type,
         raw: &'a [u8],
-    ) -> std::prelude::v1::Result<Self, Box<dyn std::error::Error + Sync + Send>> {
+    ) -> Result<Self, Box<dyn std::error::Error + Sync + Send>> {
         let mut cursor = std::io::Cursor::new(raw);
         let mut enum_value = String::new();
         cursor.read_to_string(&mut enum_value)?;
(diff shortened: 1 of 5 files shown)


src/sql/arrow_sql_gen/postgres.rs
```diff
@@ -1009,6 +1039,26 @@ impl<'a> FromSql<'a> for MoneyFromSql {
...
 
+struct EnumValueFromSql {
+    enum_value: String,
+}
+
+impl<'a> FromSql<'a> for EnumValueFromSql {
+    fn from_sql(
+        _ty: &Type,
+        raw: &'a [u8],
+    ) -> std::prelude::v1::Result<Self, Box<dyn std::error::Error + Sync + Send>> {
```
phillipleblanc, line 1050:
```suggestion
    ) -> Result<Self, Box<dyn std::error::Error + Sync + Send>> {
```

tests/postgres/common.rs
```diff
@@ -60,3 +60,38 @@ pub(super) async fn start_postgres_docker_container(
     tokio::time::sleep(std::time::Duration::from_millis(5000)).await;
     Ok(running_container)
 }
+
+fn get_pg_params_secrets(port: usize) -> HashMap<String, SecretString> {
```
phillipleblanc, line 64: Why is this duplicated from the above `get_pg_params`?

phillipleblanc: You can just use `to_secret_map`

phillipleblanc (approved):

PR #104, opened 2024-09-15
[flight] Coerce flight data into target schema if needed

ccciudatu:
Best effort column casting to avoid erroring out when the returned fields are compatible, though not exactly the same.


phillipleblanc (approved):
Makes sense to me

PR #105, opened 2024-09-16
DuckDBTableProviderFactory keeps track of opened instances

phillipleblanc:
When the `DuckDBTableProviderFactory` is now asked to create a new table, and the file path is for an instances we've already created a table for, instead of opening a new connection to the file it will reuse the existing connection. Similar for memory instances, only a single memory instance will be created instead of a new one for each table.

### DIFF
--- a/src/sql/db_connection_pool/mod.rs
+++ b/src/sql/db_connection_pool/mod.rs
@@ -1,5 +1,6 @@
 use async_trait::async_trait;
 use dbconnection::DbConnection;
+use std::sync::Arc;
 
 pub mod dbconnection;
 #[cfg(feature = "duckdb")]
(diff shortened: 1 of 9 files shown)


src/sql/db_connection_pool/mod.rs
```diff
@@ -48,3 +49,22 @@ impl From<&str> for Mode {
         }
     }
 }
+
+/// A key that uniquely identifies a database instance.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum DbInstanceKey {
+    /// The database is a file on disk, with the given path.
+    File(Arc<str>),
+    /// The database is in memory.
```
Jeadie, line 58: nit: possibly worth mentioning that we explicitly only want to allow a single in-memory db.

Jeadie (approved):

src/sql/db_connection_pool/mod.rs
phillipleblanc: I'll do this in a follow-up PR for SQLite

PR #110, opened 2024-09-18
Handle inconsistent scale in Postgres Numeric Type data

Sevenannn:
Postgres `numeric` type doesn't necessarily contain value with same scale, while arrow decimal type has a fixed scale, update the row -> arrow conversion to resolve scale difference across rows, and add a test to verify the fix.


src/sql/arrow_sql_gen/postgres.rs
```diff
@@ -78,6 +78,9 @@ pub enum Error {
     #[snafu(display("No Arrow field found for index {index}"))]
     NoArrowFieldForIndex { index: usize },
 
+    #[snafu(display("No Postgres scale information found for index {index}"))]
```
lukekim, line 81:
```suggestion
    #[snafu(display("No PostgreSQL scale found for index {index}"))]
```

src/sql/arrow_sql_gen/postgres.rs
```diff
@@ -909,7 +925,12 @@ struct BigDecimalFromSql {
 }
 
 impl BigDecimalFromSql {
-    fn to_decimal_128(&self) -> Option<i128> {
+    fn to_decimal_128_with_scale(&self, dest_scale: u16) -> Option<i128> {
+        // Resolve scale difference
```
lukekim, line 929: More meaningful comment here. "Resolve scale difference by upscaling...."?

sgrebnov (approved):

PR #115, opened 2024-09-25
Use Decimal256 instead of Decimal128 for MySQL decimal type

Sevenannn:
MySQL supports up to [65 decimal digits](https://dev.mysql.com/doc/refman/8.4/en/fixed-point-types.html#:~:text=The%20maximum%20number%20of%20digits,scale%20for%20a%20given%20column.), use decimal128 for mysql decimal type will cause error in values that have more than 38 digits. Use Decimal256 instead in this PR.

### DIFF
--- a/src/sql/arrow_sql_gen/mysql.rs
+++ b/src/sql/arrow_sql_gen/mysql.rs
@@ -1,15 +1,14 @@
 use crate::sql::arrow_sql_gen::arrow::map_data_type_to_array_builder_optional;
 use arrow::{
     array::{
-        ArrayBuilder, ArrayRef, BinaryBuilder, Date32Builder, Decimal128Builder, Float32Builder,
+        ArrayBuilder, ArrayRef, BinaryBuilder, Date32Builder, Decimal256Builder, Float32Builder,
         Float64Builder, Int16Builder, Int32Builder, Int64Builder, Int8Builder, LargeStringBuilder,
         NullBuilder, RecordBatch, RecordBatchOptions, Time64NanosecondBuilder,
         TimestampMicrosecondBuilder, UInt64Builder,
     },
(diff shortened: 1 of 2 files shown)


src/sql/arrow_sql_gen/mysql.rs
```diff
@@ -235,34 +232,66 @@ pub fn rows_to_arrow(rows: &[Row], projected_schema: &Option<SchemaRef>) -> Resu
...
+                            let Some(val) = val else {
+                                builder.append_null();
+                                continue;
+                            };
+        
+                            let val = to_decimal_256(&val);
+        
+                            builder.append_value(val);
+                        }
+                        _ => unreachable!()
```
phillipleblanc, line 293: Why is this unreachable?

phillipleblanc (approved):

src/sql/arrow_sql_gen/mysql.rs
Sevenannn: In this case, the datatype could only be either Decimal128 or Decimal256 given the mapping of `ColumnType::MYSQL_TYPE_DECIMAL` and `ColumnType::MYSQL_TYPE_NEWDECIMAL` in `map_column_to_data_type`. Would it be better to use a error here even if it's not logically unreachable?

phillipleblanc: No, I think its fine to panic for things that can't happen, since that happening would be bug that we hopefully catch in testing. But it would be nice to add a message explaining why its unreachable so that we know what our assumptions were if this was to happen.

Sevenannn: Got it, I added a comment explaining why it is unreachable

sgrebnov (approved):

PR #117, opened 2024-09-25
Fix mysql blob & text types

Sevenannn:
* Update the mapping of column types in `map_str_type_to_column_type`, since  `TINYBLOB, BLOB, MEDIUMBLOB, LONGBLOB, TINYTEXT, TEXT, MEDIUMTEXT, LONGTEXT` are all `MYSQL_TYPE_BLOB`. https://dev.mysql.com/doc/c-api/8.0/en/c-api-data-structures.html
* Use `binary` for `TINYBLOB, BLOB, MEDIUMBLOB` types, `utf8` for `TINYTEXT, TEXT, MEDIUMTEXT` types, `largebinary` for `LONGBLOB ` type, and `LargeUtf8` for `LONGTEXT ` type, according to the [MySQL type storage requirement](https://dev.mysql.com/doc/refman/8.4/en/storage-requirements.html) and [binary / non-binary attirbute for blob and text types](https://dev.mysql.com/doc/refman/8.0/en/blob.html)
* Add a test to cover the row -> arrow conversion for all `BLOB` and `TEXT` types


Sevenannn:
There's something need to be changed in this PR plz don't review yet

Sevenannn:
I wrote a bad logic determining whether to use `large` type

Sevenannn:
Updated

sgrebnov (approved):

PR #147, opened 2024-10-19
[flight] make the enforce_schema function public

ccciudatu:
Expose the (slightly enhanced) `enforce_schema` function to downstream application code.

### DIFF
--- a/src/flight.rs
+++ b/src/flight.rs
@@ -42,6 +42,8 @@ pub mod codec;
 mod exec;
 pub mod sql;
 
+pub use exec::enforce_schema;
+
 /// Generic Arrow Flight data source. Requires a [FlightDriver] that allows implementors
 /// to integrate any custom Flight RPC service by producing a [FlightMetadata] for some DDL.
 ///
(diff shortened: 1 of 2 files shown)


phillipleblanc (approved):
Thanks!

ccciudatu:
FWIW, I think this functionality is generic enough to be useful in a far broader context, beyond flight and even table-providers. My guess is there are variations of this already rewritten again and again across various projects.
@alamb do you think this would be a good fit for upstream crates like `arrow-cast` or `datafusion-common`?

alamb:
> FWIW, I think this functionality is generic enough to be useful in a far broader context, beyond flight and even table-providers. My guess is there are variations of this already rewritten again and again across various projects. @alamb do you think this would be a good fit for upstream crates like `arrow-cast` or `datafusion-common`?

I took a quick look at what `enforce_schema` does and it looks to me like it does something called "schema evolution" in other contexts

This code in DataFusion seems to do the same thing (map a batch schema to a target schema, filling in null columns / casting types when needed): 
https://docs.rs/datafusion/latest/datafusion/datasource/schema_adapter/struct.DefaultSchemaAdapterFactory.html

alamb:
The docs were a bit lackluster so I added some more here: https://github.com/apache/datafusion/pull/13063

PR #151, opened 2024-10-23
Implement native schema inference for PostgreSQL

### LINKED ISSUE #127: Query postgres information_schema to construct datafusion table schema
Trying to query an emtpy table results in a schema error, and I noticed that an empty table in postgres did not have any schema when mapped via a TableProvider, and after some digging saw that the schema is determined via a `SELECT * FROM {table} LIMIT 1` in https://github.com/datafusion-contrib/datafusion-table-providers/blob/5bbbb32704ad8bfd534c7fa9b6651df03c0ff6bd/src/sql/db_connection_pool/dbconnection/postgresconn.rs#L85  so no data means no schema. 

A potential improvement that would allow querying empty tables is to build the schema by querying postgres's information_schema and mapping it to arrow/df schema. 

I didn't look into other implementations to determine if other databases share this issue.

phillipleblanc:
Closes #127

This PR implements native schema inference for PostgreSQL using a combination of postgres native system tables and `information_schema`. It handles all primitive types, array types, complex types and enums. It also handles quite a few types that the row conversion itself doesn't support yet - but those can be added over time (i.e. `TSVECTOR`, `INET`, `CIDR`, `MACADDR`, `UUID`, `JSON`, `JSONB`, `XML`).

This allows the schema to be retrieved even for tables without any rows.


phillipleblanc:
I plan to cherry-pick this commit to `main` once I verify it on this branch

sgrebnov (approved):

phillipleblanc:
Will need to investigate these upstream errors before merging this PR: https://github.com/spiceai/spiceai/actions/runs/11484314326/job/31961830960

Sevenannn (approved):

PR #158, opened 2024-10-30
SQLite: rely on projected schema to convert records 

sgrebnov:
When the first returned row contains a `NULL` value, the column type is inferred as `NULL`, causing all subsequent rows' values in that column to be converted to `NULL`.

This PR addresses the issue by using the projected schema when converting records. It also avoids unnecessary conversion during the `try_cast_to` step for primitive types, ensuring they are converted directly to the target types.

This resolves incorrect results produced with SQLite acceleration for TPC-DS queries Q51 and Q59.

Note:
 - This PR does not change the behavior related to Decimal conversion (continue to use `Float64` that then will be converted to `Decimal` with `try_cast_to`). There will be a follow up PR to improve this as well.

### DIFF
--- a/src/sql/arrow_sql_gen/sqlite.rs
+++ b/src/sql/arrow_sql_gen/sqlite.rs
@@ -20,12 +20,22 @@ use crate::sql::arrow_sql_gen::arrow::map_data_type_to_array_builder;
 use arrow::array::ArrayBuilder;
 use arrow::array::ArrayRef;
 use arrow::array::BinaryBuilder;
+use arrow::array::BooleanBuilder;
+use arrow::array::Float32Builder;
 use arrow::array::Float64Builder;
+use arrow::array::Int16Builder;
+use arrow::array::Int32Builder;
 use arrow::array::Int64Builder;
+use arrow::array::Int8Builder;
+use arrow::array::LargeStringBuilder;
 use arrow::array::NullBuilder;
 use arrow::array::RecordBatch;
 use arrow::array::RecordBatchOptions;
 use arrow::array::StringBuilder;
+use arrow::array::UInt16Builder;
+use arrow::array::UInt32Builder;
+use arrow::array::UInt64Builder;
+use arrow::array::UInt8Builder;
 use arrow::datatypes::DataType;
 use arrow::datatypes::Field;
 use arrow::datatypes::Schema;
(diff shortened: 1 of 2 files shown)


src/sql/arrow_sql_gen/sqlite.rs
```diff
@@ -79,42 +93,22 @@ pub fn rows_to_arrow(mut rows: Rows, num_cols: usize, projected_schema: Option<S
...
-            // Note: The same column may contain both integer and floating point values. 
-            // Reading values as Float is safe even if the value is stored as an integer.
-            // Refer to the rusqlite type handling documentation for more details:
-            // https://github.com/rusqlite/rusqlite/blob/95680270eca6f405fb51f5fbe6a214aac5fdce58/src/types/mod.rs#L21C1-L22C75
-            //
-            // `REAL` to integer: always returns an [`Error::InvalidColumnType`](crate::Error::InvalidColumnType) error.
-            // `INTEGER` to float: casts using `as` operator. Never fails.
-            // `REAL` to float: casts using `as` operator. Never fails.
-
-            if column_type == Type::Integer {
```
sgrebnov, line 94: This is no longer actual as we rely on projected schema.

src/sql/arrow_sql_gen/sqlite.rs
```diff
@@ -130,6 +124,32 @@ pub fn rows_to_arrow(mut rows: Rows, num_cols: usize, projected_schema: Option<S
...
+        DataType::UInt32 => DataType::UInt32,
+        DataType::UInt64 => DataType::UInt64,
+        DataType::Boolean => DataType::Boolean,
+        DataType::Float16 => DataType::Float16,
+        DataType::Float32 => DataType::Float32,
+        DataType::Float64 => DataType::Float64,
+        DataType::Utf8 => DataType::Utf8,
+        DataType::LargeUtf8 => DataType::LargeUtf8,
+        DataType::Binary | DataType::LargeBinary | DataType::FixedSizeBinary(_) => DataType::Binary,
+        DataType::Decimal128(_, _) | DataType::Decimal256(_, _) => DataType::Float64,
```
sgrebnov, line 145: This is existing behavior. Will be improved to directly convert to Decimal by separate PR

src/sql/arrow_sql_gen/sqlite.rs
```diff
@@ -130,6 +124,32 @@ pub fn rows_to_arrow(mut rows: Rows, num_cols: usize, projected_schema: Option<S
...
+        DataType::Duration(_) => DataType::Int64,
+
+        // Timestamp, Date32, Date64, Time32, Time64, List, Struct, Union, Dictionary, Map
+        _ => DataType::Utf8,
```
sgrebnov, line 149: All those types are stored in SQLite as text, they are converted to actual type during `try_cast_to`

tests/sqlite/mod.rs
```diff
@@ -55,32 +55,50 @@ async fn arrow_sqlite_round_trip(
...
-    let sql = format!("SELECT * FROM {table_name}");
-    let df = ctx
-        .sql(&sql)
-        .await
-        .expect("DataFrame should be created from query");
 
-    let record_batch = df.collect().await.expect("RecordBatch should be collected");
+    // Perform the test twice: first, simulate a request without a known schema;
+    // then, test result conversion with a known projected schema (matching the test RecordBatch).
+    for projected_schema in vec![None, Some(arrow_record.schema())] {
```
sgrebnov, line 61: The second case with the projected schema simulates the actual behavior, as the schema is inferred by the connector.

tests/sqlite/mod.rs
```diff
@@ -92,8 +110,7 @@ async fn arrow_sqlite_round_trip(
 #[case::timestamp(get_arrow_timestamp_record_batch(), "timestamp")]
 #[case::date(get_arrow_date_record_batch(), "date")]
 #[case::struct_type(get_arrow_struct_record_batch(), "struct")]
-// SQLite only supports up to 16 precision for decimal through REAL type.
-#[case::decimal(get_sqlite_arrow_decimal_record_batch(), "decimal")]
+#[case::decimal(get_arrow_decimal_record_batch(), "decimal")]
```
sgrebnov, line 113: Limitation is no longer actual

peasee (approved):

peasee (approved):

PR #166, opened 2024-11-12
Fix `MySQLConnection::get_schema` for uppercase `TableReference`

Jeadie:
For MySQL servers with table names (or schema, catalogs), `get_schema` does not handle uppercases correctly.
 - `table_reference.to_quoted_string()` uses the `DefaultDialect` which has a different `identifier_quote_style` than `MySqlDialect`.

### DIFF
--- a/src/sql/db_connection_pool/dbconnection/mysqlconn.rs
+++ b/src/sql/db_connection_pool/dbconnection/mysqlconn.rs
@@ -45,7 +45,7 @@ pub struct MySQLConnection {
 }
 
 impl MySQLConnection {
-    /// Create a [`TableReference`] in a manner that properly handles the unique quote style for MySQL.
+    /// Create a [`TableReference`] in a manner that properly handles the unique quote style of MySQL.
     ///
     /// [`TableReference::from`] uses `DefaultDialect` and therefore gets quoting incorrect.
     fn to_mysql_quoted_string(tbl: &TableReference) -> String {
--- a/src/sql/db_connection_pool/dbconnection/mysqlconn.rs
+++ b/src/sql/db_connection_pool/dbconnection/mysqlconn.rs
@@ -53,7 +53,13 @@ impl MySQLConnection {
         let tbl_name = [tbl.catalog(), tbl.schema(), Some(tbl.table())]
             .into_iter()
             .filter_map(|part| part)
-            .map(|part| format!("{quote}{part}{quote}", quote = q))
+            .map(|part| {
+                if part.starts_with(quote) && part.ends_with(quote) {
+                    part.to_string()
+                } else {
+                    format!("{quote}{part}{quote}", quote = q)
+                }
+            })
             .collect::<Vec<_>>()
             .join(".");
 
--- a/src/sql/db_connection_pool/dbconnection/mysqlconn.rs
+++ b/src/sql/db_connection_pool/dbconnection/mysqlconn.rs
@@ -7,6 +7,7 @@ use async_stream::stream;
 use datafusion::error::DataFusionError;
 use datafusion::execution::SendableRecordBatchStream;
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
+use datafusion::sql::unparser::dialect::{Dialect, MySqlDialect};
 use datafusion::sql::TableReference;
 use futures::lock::Mutex;
 use futures::{stream, StreamExt};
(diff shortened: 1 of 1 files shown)


src/sql/db_connection_pool/dbconnection/mysqlconn.rs
```diff
@@ -43,6 +44,26 @@ pub struct MySQLConnection {
     pub conn: Arc<Mutex<Conn>>,
 }
 
+impl MySQLConnection {
+    fn column_meta_query_for(tbl: &TableReference) -> String {
+        let q = MySqlDialect {}
+            .identifier_quote_style("") // parameter unimportant for `MySqlDialect`.
```
sgrebnov, line 50: Why we override MySQL default `identifier_quote_style`?

```
impl Dialect for MySqlDialect {
    fn identifier_quote_style(&self, _: &str) -> Option<char> {
        Some('`')
    }
```

sgrebnov: Might be better to just add utility function with own implementation w/o using MySqlDialect?

Jeadie: Fixed part 2. For part 1. wanted to keep it non-hardcoded in case the pattern needs to be used elsewhere.

Jeadie: Also `column_meta_query_for` is a utility function, just under the `MySQLConnection` implementation. Is there a better spot?

sgrebnov: For readability, future modifications and other contributors I think it might be beneficial to have it as a separate `to_mysql_quoted_string` fn with a note why `table_reference.to_quoted_string()` can't be used

src/sql/db_connection_pool/dbconnection/mysqlconn.rs
```diff
@@ -72,13 +99,10 @@ impl<'a> AsyncDbConnection<Conn, &'a (dyn ToValue + Sync)> for MySQLConnection {
...
 
-        // we don't use SELECT COLUMN_NAME, DATA_TYPE FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = '{}' AND TABLE_SCHEMA = '{}'
-        // as table_reference don't always have schema specified so we need to extract schema/db name from connection properties
-        // to ensure we are querying information for correct table
-        let columns_meta_query =
-            format!("SHOW COLUMNS FROM {}", table_reference.to_quoted_string());
-
-        let columns_meta: Vec<Row> = match conn.exec(&columns_meta_query, Params::Empty).await {
+        let columns_meta: Vec<Row> = match conn
+            .exec(Self::column_meta_query_for(table_reference), Params::Empty)
```
Sevenannn, line 103: Instead of building query as string, it would be safer to use prepared statement and pass the table name as value, something like `tx.exec("INSERT INTO tmp (a) VALUES (?)", ("foo",))?;`

src/sql/db_connection_pool/dbconnection/mysqlconn.rs
Jeadie: This is wise. I like it.

src/sql/db_connection_pool/dbconnection/mysqlconn.rs
Sevenannn: The current approach could be vulnerable to SQLi, user can pass in name like this: 
```
fake_name`; DELETE TABLE secrets;--
```
and it will be passed in as query param to perform bad operations
```
SHOW COLUMNS FROM `catalog`.`schema`.`fake_name`; DELETE TABLE secrets;--`
```

Jeadie: From what I found, `SHOW` doesn't support prepared statements in that manner.

sgrebnov: @Sevenannn  - good catch. This is DF reference implementation - should we just follow the same approach.



```
/// Wraps identifier string in double quotes, escaping any double quotes in
/// the identifier by replacing it with two double quotes
///
/// e.g. identifier `tab.le"name` becomes `"tab.le""name"`
pub fn quote_identifier(s: &str) -> Cow<str> {
    if needs_quotes(s) {
        Cow::Owned(format!("\"{}\"", s.replace('"', "\"\"")))
    } else {
        Cow::Borrowed(s)
    }
}

/// returns true if this identifier needs quotes
fn needs_quotes(s: &str) -> bool {
    let mut chars = s.chars();

    // first char can not be a number unless escaped
    if let Some(first_char) = chars.next() {
        if !(first_char.is_ascii_lowercase() || first_char == '_') {
            return true;
        }
    }

    !chars.all(|c| c.is_ascii_lowercase() || c.is_ascii_digit() || c == '_')
}
```

```
pub fn to_quoted_string(&self) -> String {
        match self {
            TableReference::Bare { table } => quote_identifier(table).to_string(),
            TableReference::Partial { schema, table } => {
                format!("{}.{}", quote_identifier(schema), quote_identifier(table))
            }
            TableReference::Full {
                catalog,
                schema,
                table,
            } => format!(
                "{}.{}.{}",
                quote_identifier(catalog),
                quote_identifier(schema),
                quote_identifier(table)
            ),
        }
    }
```

Jeadie: I don't think we need all of this?

Sevenannn: The `parse_multipart_identifier` used on the Spice side seem to be pretty strict on parsing the catalog / schema / table name as it doesn't allow extraneous tokens so seems that this wouldn't cause SQLi in Spice. I wonder if the method should be updated here for people merely uses datafusion-table-provider repo.

Jeadie: Yes, but it doesn't work for `SHOW COLUMNS` in MySQL.

phillipleblanc (approved):

PR #168, opened 2024-11-12
[flight] Enhanced customizability

ccciudatu:
Expose config options to provide full control over the protocol details, allowing users to fine-tune the client behaviour for their specific service.


ccciudatu:
Looks like datafusion-federation 0.3.1 breaks both tests and clippy. Added a commit to pin it to the previous version.

phillipleblanc (approved):
Looks good - I think the newer datafusion-federation crate upgraded the DataFusion version to 43. I will backport my DF upgrade to main and handle the upgrade then.

PR #173, opened 2024-11-15
Fix MySQL timestamp conversion when running via `MySQLSQLExec`

sgrebnov:
Fix MySQL timestamp conversion:  it does not support `TO_TIMESTAMP(..)` and requires `FROM_UNIXTIME(..)`

This fixes error below when using append mode with MySQL data connector
>Unable to query arrow: Server error: `ERROR 42000 (1305): FUNCTION spice_demo.TO_TIMESTAMP does not exist'

### DIFF
--- a/src/sql/sql_provider_datafusion/expr.rs
+++ b/src/sql/sql_provider_datafusion/expr.rs
@@ -113,7 +113,9 @@ pub fn to_sql_with_engine(expr: &Expr, engine: Option<Engine>) -> Result<String>
                         Ok(format!("TO_TIMESTAMP({})", *value as f64 / 1_000_000.0))
                     }
                 }
-                Some(Engine::MySQL) => Ok(format!("TIMESTAMP({})", value / 1_000_000)),
+                Some(Engine::MySQL) => {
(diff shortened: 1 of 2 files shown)


src/sql/sql_provider_datafusion/expr.rs
```diff
@@ -112,6 +113,7 @@ pub fn to_sql_with_engine(expr: &Expr, engine: Option<Engine>) -> Result<String>
                         Ok(format!("TO_TIMESTAMP({})", *value as f64 / 1_000_000.0))
                     }
                 }
+                Some(Engine::MySQL) => Ok(format!("TIMESTAMP({})", value / 1_000_000)),
```
sgrebnov, line 116: MySQL does not support `TO_TIMESTAMP(..)` and requires `TIMESTAMP(..)`

Sevenannn, line 116: `value` should be converted to f64 to keep the precision after seconds in timestamp.
Example: 
image

sgrebnov: Updated

Sevenannn (approved):

PR #180, opened 2024-11-21
Fix MySQL errors

Sevenannn:
- Remove text message from `dbconnection::Error::UnableToGetSchema`. This error is widely used within the `get_schema` function itself, which really should just bubble up the internal error instead of wrapping duplicated `Unable to get schema:`, error message `Unable to get schema:` should be set and handled by caller of the `get_schema` function, for example, in `sql_provider_datafusion::Error`.
- Update the column type message to include column name information.
- Add file issue guide for catastrophic error caused by bugs in datafusion-table-providers, and bugs caused by things currently unsupported in the repo 
- Add MySQL manual reference for MySQL query errors

Besides changes for MySQL, this PR also include changes for instructing users to file an issue for DuckDB type related errors.

Before:
image
After:
image

### DIFF
--- a/src/sql/db_connection_pool/dbconnection/mysqlconn.rs
+++ b/src/sql/db_connection_pool/dbconnection/mysqlconn.rs
@@ -27,7 +27,7 @@ pub enum Error {
     #[snafu(display("Failed to convert query result to Arrow: {source}.\nThis was likely caused by a bug in the DataFusion Table Providers code, which you can report here: https://github.com/datafusion-contrib/datafusion-table-providers/issues"))]
     ConversionError { source: arrow_sql_gen::mysql::Error },
 
-    #[snafu(display("Unable to get MySQL query result stream"))]
+    #[snafu(display("Unable to get MySQL query result stream.\nReport a bug to request support: https://github.com/datafusion-contrib/datafusion-table-providers/issues"))]
     QueryResultStreamError {},
 
     #[snafu(display("The field '{column_name}' has an unsupported data type: {data_type}.\nReport a bug to request support for this data type: https://github.com/datafusion-contrib/datafusion-table-providers/issues"))]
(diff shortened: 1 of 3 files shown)


src/sql/db_connection_pool/dbconnection/mysqlconn.rs
```diff
@@ -30,13 +30,16 @@ pub enum Error {
...
-    #[snafu(display("Unsupported column data type: {data_type}"))]
-    UnsupportedDataTypeError { data_type: String },
+    #[snafu(display("The field '{column_name}' contains unsupported data type: {data_type}. \n Please file an issue https://github.com/datafusion-contrib/datafusion-table-providers"))]
+    UnsupportedDataTypeError {
+        column_name: String,
+        data_type: String,
+    },
 
-    #[snafu(display("Unable to extract precision and scale from type: {data_type}"))]
+    #[snafu(display("Unable to extract precision and scale from type: {data_type}. \n This was likely caused by a bug in DataFusion Table Providers code, please file an issue https://github.com/datafusion-contrib/datafusion-table-providers"))]
```
peasee, line 39:
```suggestion
    #[snafu(display("Unable to extract precision and scale from type: {data_type}.\nThis was likely caused by a bug in the DataFusion Table Providers code, which you can report here: https://github.com/datafusion-contrib/datafusion-table-providers/issues"))]
```

src/sql/db_connection_pool/dbconnection/mysqlconn.rs
```diff
@@ -30,13 +30,16 @@ pub enum Error {
...
     UnableToGetDecimalPrecisionAndScale { data_type: String },
 
-    #[snafu(display("Field '{field}' is missing"))]
+    #[snafu(display("Field '{field}' is missing. \n This was likely caused by a bug in DataFusion Table Providers code, please file an issue https://github.com/datafusion-contrib/datafusion-table-providers"))]
```
peasee, line 42:
```suggestion
    #[snafu(display("Failed to find the field '{field}'.\nThis was likely caused by a bug in the DataFusion Table Providers code, which you can report here: https://github.com/datafusion-contrib/datafusion-table-providers/issues"))]
```

peasee (approved):

PR #201, opened 2024-12-20
Allow overriding the default DuckDB dialect

sgrebnov:
Allow overriding the default DuckDB dialect


src/duckdb/sql_table.rs
```diff
@@ -44,6 +45,7 @@ impl<T, P> DuckDBTable<T, P> {
         schema: impl Into<SchemaRef>,
         table_reference: impl Into<TableReference>,
         table_functions: Option<HashMap<String, String>>,
+        dialect: Option<Arc<dyn Dialect + Send + Sync>>,
```
sgrebnov, line 48: To simplify DuckDBTable usage dialect parameter is optional and fallback to default one if not specified

phillipleblanc (approved):

PR #219, opened 2025-01-13
chore: upgrade datafusion and and some other crates

xhwhis:
@phillipleblanc Can you update a new version to crates.io?


phillipleblanc (approved):
Thanks!

PR #226, opened 2025-01-15
fix: Use Unparser for expr to sql

peasee:
* Cherry picks a couple changes from `main` that include helpers to get dialect by engine, and `filters_to_sql` updates to use the DataFusion Unparser
* Updates the `sql()` function of the `SqlExec` to use the DataFusion Unparser instead of `to_sql_with_engine`

### DIFF
--- a/src/sql/sql_provider_datafusion/expr.rs
+++ b/src/sql/sql_provider_datafusion/expr.rs
@@ -1,7 +1,12 @@
+use std::sync::Arc;
+
 use bigdecimal::{num_bigint::BigInt, BigDecimal};
 use datafusion::{
     logical_expr::{Cast, Expr},
     scalar::ScalarValue,
+    sql::unparser::dialect::{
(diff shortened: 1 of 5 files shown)


src/sql/sql_provider_datafusion/expr.rs
```diff
@@ -236,6 +253,16 @@ fn handle_cast(cast: &Cast, engine: Option<Engine>, expr: &Expr) -> Result<Strin
...
         },
+        arrow::datatypes::DataType::Int64 => match engine {
+            Some(Engine::DuckDB | Engine::SQLite | Engine::Postgres) => Ok(format!(
+                "CAST({} AS BIGINT)",
+                to_sql_with_engine(&cast.expr, engine)?,
+            )),
+            _ => Ok(format!(
+                "CAST({} AS INT64)",
+                to_sql_with_engine(&cast.expr, engine)?,
+            )),
```
phillipleblanc, line 264: I think we can just always cast this as BIGINT?

phillipleblanc (approved):

PR #231, opened 2025-02-04
fix: Postgres LargeUtf8 is equivalent to Utf8

peasee:
* In Postgres, the largest string that can be stored is 1GB (VARCHAR). This is smaller than the largest string a `Utf8` in DataFusion can represent, so for the purposes of type equivalency `LargeUtf8` and `Utf8` are equal for Postgres.

### DIFF
--- a/src/postgres/write.rs
+++ b/src/postgres/write.rs
@@ -1,7 +1,7 @@
 use std::{any::Any, fmt, sync::Arc};
 
 use arrow::datatypes::SchemaRef;
-use arrow_schema::{DataType, Schema};
+use arrow_schema::{DataType, Field, Schema};
 use async_trait::async_trait;
 use datafusion::{
     catalog::Session,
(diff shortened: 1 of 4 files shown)


src/sql/arrow_sql_gen/statement.rs
peasee, line 1: this file got formatted

src/postgres/write.rs
```diff
@@ -141,10 +142,24 @@ impl DataSink for PostgresDataSink {
...
+                        && DFSchema::datatype_is_semantically_equal(
+                            pg_field.data_type(), // for the purposes of PostgreSQL, LargeUtf8 is equivalent to Utf8
+                            // because Postgres physically cannot store anything larger than 1Gb in text (VARCHAR)
+                            if batch_field.data_type() == &DataType::LargeUtf8 {
+                                &DataType::Utf8
+                            } else {
+                                batch_field.data_type()
+                            },
+                        )
+                })
```
phillipleblanc, line 162: This normalizes the `batch_field` type from `LargeUtf8` to `Utf8` - but what if the `pg_field` type is `LargeUtf8`? I think a better approach would be to collect all of the types from each field, and normalize them, then run through the existing `equivalent_names_and_types` check.

phillipleblanc (approved):

PR #260, opened 2025-03-15
Add primary keys mismatch verification for DuckDB table creation

sgrebnov:
>2025-03-15T01:14:39.619510Z  WARN datafusion_table_providers::duckdb: The table 'issues' has unexpected primary key(s) not defined in the configuration: "id".
2025-03-15T01:14:39.619541Z  WARN datafusion_table_providers::duckdb: Schema mismatch detected:
The local table definition for 'issues' in database '/Users/sg/spice/spiceai/.spice/data/accelerated_duckdb.db' does not match the expected configuration.
To fix this, drop the existing local copy. A new table with the correct schema will be automatically created upon the next access.

image

### DIFF
--- a/src/duckdb.rs
+++ b/src/duckdb.rs
@@ -1,4 +1,5 @@
 use crate::sql::sql_provider_datafusion;
+use crate::util::constraints::get_primary_keys_from_constraints;
 use crate::util::{
     self,
     column_reference::{self, ColumnReference},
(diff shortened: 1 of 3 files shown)


src/duckdb.rs
```diff
@@ -346,6 +352,21 @@ impl TableProviderFactory for DuckDBTableProviderFactory {
...
             .map_err(to_datafusion_error)?;
 
+        // If the table is already created, we don't create it again and don't apply primary keys.
+        // We need to verify that the table was created with the correct primary keys.
+        if !duckdb
+            .verify_primary_keys_match()
+            .await
+            .map_err(to_datafusion_error)?
+        {
+            tracing::warn!(
```
sgrebnov, line 362: As a next step we might want to add more control similar to [unsupported_type_action](https://spiceai.org/docs/reference/spicepod/datasets#unsupported_type_action)

lukekim (approved):

PR #194, opened 2024-12-10
Initial commit for python table providers

timsaucer:
This PR adds support for table providers with `datafusion-python` versions 43.1.0 and later. Included is an example of using sqlite and registering the tables in that database with the datafusion session context.

Still todo is to flush out the other table providers besides sqlite and to make sure all of the recommended options have been exposed.

Still todo is to add unit tests.

PR #264, opened 2025-03-16
feat: Add python FFI interface and expose sqlite and duckdb table providers

CrystalZhou0529:
Expose table providers in Python via FFI

- Forked from #194 and fixed python dependencies

## Summary
1. Created python subcrate for python-related development
2. Reorganized Cargo dependency and share common dependencies (arrow, datafusion, duckdb) in workspace cargo.toml
3. Implemented sqlite and duckdb table provider interfaces and tested in examples
4. Updated `duckdbconn.rs` to start new tokio runtime when triggered from PyO3
5. Updated Github CICD testing command

## Testing
- Verified that all existing tests in Rust have passed
- Verified that Python examples can execute correctly via the following script:
```sh
cd python
maturin develop
cd examples
python3 duckdb_demo.py
```

## Follow-up works
- Improve duckdb enum interface (ongoing)

The rest will be left for future works: #279 
- Python ODBC table provider + example
- Python Postgres table provider + example
- Python MySQL table provider + example
- Integration testing
- Finalize Python interface design, i.e. `new_memory()` or `new(":memory:")`


core/src/sql/db_connection_pool/dbconnection/duckdbconn.rs
```diff
@@ -281,6 +282,13 @@ impl DbConnection<r2d2::PooledConnection<DuckdbConnectionManager>, DuckDBParamet
     }
 }
 
+fn get_tokio_runtime() -> &'static Runtime {
+    // TODO: this function is a repetition of python/src/utils.rs::get_tokio_runtime.
+    // Think about how to refactor it
+    static RUNTIME: OnceLock<Runtime> = OnceLock::new();
+    RUNTIME.get_or_init(|| Runtime::new().expect("Failed to create Tokio runtime"))
+}
+
```
CrystalZhou0529, line 291: Discussed in below. This function should be moved to somewhere else, preferably to the Python side.

python/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+[lib]
+name = "datafusion_table_providers"
+crate-type = ["cdylib"]
+doc = false
+
+[dependencies]
+arrow = { workspace = true }
+datafusion = { workspace = true, features = ["pyarrow"] }
+datafusion-ffi = { workspace = true }
+datafusion-table-providers = { workspace = true, features = ["sqlite", "duckdb"] }
```
CrystalZhou0529, line 19: Should we enable all features by default or have feature flags to control them?

python/examples/duckdb_demo.py
```diff
@@ -0,0 +1,11 @@
...
+from datafusion_table_providers import duckdb
+
+ctx = SessionContext()
+pool = duckdb.DuckDBTableFactory("../../core/examples/duckdb_example.db", "READ_ONLY")
+tables = pool.tables()
+
+for t in tables:
+    ctx.register_table_provider(t, pool.get_table(t))
+    print("Checking table:", t)
+    ctx.table(t).show()
```
CrystalZhou0529, line 11: When testing, I encountered an error at this line for uninitalized tokio runtime. I managed to resolve it by adding `get_tokio_runtime` to [duckdbconn.rs](https://github.com/datafusion-contrib/datafusion-table-providers/blob/ad3d352240e203fc03bc2049c5f146a827b12231/core/src/sql/db_connection_pool/dbconnection/duckdbconn.rs#L285) and wrapping `query_arrow` with it. However, this is not an ideal workaround because we only need to manually start tokio from PyO3, and we would prefer not to modify the existing rust code.
I presume that sqlite implementation does not run into the same issue because it supports async connection, whereas duckdb only supports synchronous connection.

My intuition is, we need to make some changes in `datafusion-python` and move `get_tokio_runtime` to it. In this way, we can wrap all the necessary functions with `get_tokio_runtime`.

python/python/datafusion_table_providers/duckdb.py
```diff
@@ -0,0 +1,28 @@
...
+            path: Memory or file location
+            access_mode: Access mode configuration
+        """
+        # TODO: think about the interface, restrict invalid combination of input
+        # arguments, for example, if path is memory, then access_mode should not be
+        # specified.
+        if path == ":memory:" or path == "":
+            self._raw = _internal.duckdb.RawDuckDBTableFactory.new_memory()
+        else:
+            self._raw = _internal.duckdb.RawDuckDBTableFactory.new_file(path, access_mode)
```
CrystalZhou0529, line 22: Will consult DuckDB interface to get more insights. Also looking for a second thought on the interface design because I know the current one is bad :(
I would like a way to restrict the interface to `new_memory()` or `new_file(path, access_mode)` but overloading is not supported in python constructor.

python/src/duckdb.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        Ok(Self {
+            factory: DuckDBTableFactory::new(Arc::clone(&pool)),
+            pool,
+        })
+    }
+
+    #[staticmethod]
+    #[pyo3(signature = (path, access_mode))]
+    pub fn new_file(path: &str, access_mode: &str) -> PyResult<Self> {
```
CrystalZhou0529, line 36: What type should we define `access_mode` to? We need `duckdb::AccessMode`, but it did not implement necessary traits for PyO3 to use. Our options are:
1. Use string and do conversion via `AccessMode::from_str`. This is easier but less safe
2. Define a copy of `AccessMode` enum and implement necessary traits. Also manually do the conversion. This is more complex but safer.
3. (looking for thoughts)

python/examples/duckdb_demo.py
CrystalZhou0529: @timsaucer Here are some thoughts on the tokio runtime issue. Please take a look and share some insights with me if possible!

core/Cargo.toml
```diff
@@ -0,0 +1,165 @@
...
+[dependencies]
+arrow = { workspace = true }
+arrow-array = { version = "54.2.1", optional = true }
+arrow-flight = { version = "54.2.1", optional = true, features = [
+  "flight-sql-experimental",
+  "tls",
+] }
+arrow-schema = { version = "54.2.1", optional = true, features = ["serde"] }
+arrow-json = "54.2.1"
+arrow-odbc = { version = "=15.1.1", optional = true }
```
timsaucer, line 19: I recommend we put all of the arrow dependencies in one place - the workspace Cargo.toml. That way when someone goes to update them, they don't miss one easily.

core/Cargo.toml
```diff
@@ -0,0 +1,165 @@
...
+bb8 = { version = "0.9", optional = true }
+bb8-postgres = { version = "0.9", optional = true }
+bigdecimal = "0.4"
+byteorder = "1.5.0"
+bytes = { version = "1.7.1", optional = true }
+byte-unit = { version = "5.1.4", optional = true }
+chrono = "0.4"
+dashmap = "6.1.0"
+datafusion = { version = "45", default-features = false }
+datafusion-expr = { version = "45", optional = true }
```
timsaucer, line 32: Similar comment as above, recommend putting all datafusion dependencies in workspace Cargo.toml

python/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "datafusion-table-providers-python"
+version = { workspace = true }
+readme = { workspace = true }
+edition = { workspace = true }
+repository = { workspace = true }
+license = { workspace = true }
+description = { workspace = true }
```
timsaucer, line 8: One thing I'm not sure about - do we want this crate (the python crate) published to crates.io or not? @phillipleblanc do you have an opinion? I don't expect there would be a reason for someone to need it as a dependency.

python/examples/duckdb_demo.py
timsaucer: I think we need to make the change in the rust code of this crate. The tokio runtime is an *extremely* tricky topic. We have multiple runtimes running around because we have different libraries and we cannot share that runtime across the ffi boundary. I'll see if I can reproduce the problem and where I think it should get exposed.

python/pyproject.toml
```diff
@@ -0,0 +1,59 @@
...
+    "Operating System :: MacOS",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: POSIX :: Linux",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python",
+    "Programming Language :: Rust",
+]
+dependencies = ["datafusion>=43.0.0"]
```
timsaucer, line 27: Can you double check which version of datafusion-python first has the tokio runtime? I *think* it might be 44. It will be important.

python/python/datafusion_table_providers/duckdb.py
timsaucer: I suggest we focus on getting some immediate functionality now but do not yet publish to pypi until we are happy with the interfaces.

timsaucer (commented):
Overall, this looks very nice. I suggest we merge this in but do not yet publish to pypi. There should be no breaking changes to the rust code, so anyone who depends on it should be okay. I think we will need to bump the minor version number if you haven't done so (I didn't check).

@phillipleblanc Would you be comfortable if we merged in partially complete work on the python side so we can have a series of smaller PRs until we're ready to publish?

python/Cargo.toml
phillipleblanc: Let's not publish it for now - if someone needs it later we can always publish it then.

phillipleblanc:
> Overall, this looks very nice. I suggest we merge this in but do not yet publish to pypi. There should be no breaking changes to the rust code, so anyone who depends on it should be okay. I think we will need to bump the minor version number if you haven't done so (I didn't check).
> 
> @phillipleblanc Would you be comfortable if we merged in partially complete work on the python side so we can have a series of smaller PRs until we're ready to publish?

Yes, that works for me - just mark this PR as ready to review and tag me when you are ready to merge. Thanks for working on this!

python/pyproject.toml
CrystalZhou0529: If I understand correctly it is in 43? Ref: https://github.com/apache/datafusion-python/blob/b8dd97bc8eefcfecfa8dcc864c4898c654b236a9/dev/changelog/pre-43.0.0.md?plain=1#L53

timsaucer: This is the commit where the tokio runtimes were added on the FFI side: https://github.com/apache/datafusion/pull/13937 and that is in datafusion 45. I think we need that version as the minimum.

python/Cargo.toml
timsaucer: Great! @CrystalZhou0529 we want a `publish = false` line like in this example from datafusion: https://github.com/apache/datafusion/blob/main/datafusion-examples/Cargo.toml

CrystalZhou0529: Sounds good!

python/pyproject.toml
CrystalZhou0529: Fixed!

CrystalZhou0529:
@timsaucer @phillipleblanc This PR is ready for review!

core/Cargo.toml
```diff
@@ -0,0 +1,165 @@
+[package]
+name = "datafusion-table-providers"
+version = "0.3.0"
```
timsaucer, line 3: I recommend pulling as much of this from the workspace as possible

```
version = { workspace = true }
edition = { workspace = true }
homepage = { workspace = true }
repository = { workspace = true }
```

and so on

python/python/datafusion_table_providers/sqlite.py
```diff
@@ -0,0 +1,20 @@
...
+
+class SqliteTableFactory:
+
+    def __init__(self, path: str, mode: str, busy_timeout_s: float, attach_databases: Optional[List[str]] = None) -> None:
+        # TODO: fix documentation
+        """Create a new :py:class:`SessionConfig` with the given configuration options.
+
+        Args:
+            config_options: Configuration options.
+        """
```
timsaucer, line 13: Stale comment?

timsaucer (approved):
I have just a couple of smaller suggestions. This feels like a huge leap forward in making these table providers accessible. I do recommend updating the title of this PR.

Also coming in DF 47, whenever it releases, we will have FFI catalog providers. Some of these interfaces might work well to support that as well. But that is future work. For now, this is excellent. Well done.

python/python/datafusion_table_providers/sqlite.py
CrystalZhou0529: Updated.

phillipleblanc (approved):

phillipleblanc:
Thank you @CrystalZhou0529 and @timsaucer!

PR #275, opened 2025-03-19
Allow connection pool size configuration for duckdb connection pool

Sevenannn:
Allow the configuration of connection pool size for duckdb connection pool through `get_or_init_memory_instance` and `get_or_init_file_instance`. This is useful when the duckdb instance will be used under high concurrency.

`min_idle` is only configured when `max_size` is presented. The min_idle uses the r2d2 duckdb `max_size` default: 10. The max_size takes the `max(min_idle, connection_pool_size)`.

### DIFF
--- a/src/duckdb.rs
+++ b/src/duckdb.rs
@@ -275,7 +275,7 @@ impl DuckDBTableProviderFactory {
             return Ok(instance.clone());
         }
 
-        let mut pool_builder = DuckDbConnectionPoolBuilder::default().memory();
+        let mut pool_builder = DuckDbConnectionPoolBuilder::memory();
 
         if max_size.is_some() {
             pool_builder = pool_builder.with_max_size(max_size);
@@ -315,10 +315,8 @@ impl DuckDBTableProviderFactory {
             AccessMode::Automatic => AccessMode::Automatic,
         };
 
-        let mut pool_builder = DuckDbConnectionPoolBuilder::default()
-            .with_access_mode(access_mode)
-            .with_file_path(&db_path)
-            .file();
+        let mut pool_builder =
+            DuckDbConnectionPoolBuilder::file(&db_path).with_access_mode(access_mode);
 
         if max_size.is_some() {
             pool_builder = pool_builder.with_max_size(max_size);
--- a/src/duckdb.rs
+++ b/src/duckdb.rs
@@ -255,8 +255,7 @@ impl DuckDBTableProviderFactory {
 
     pub async fn get_or_init_memory_instance(&self) -> Result<DuckDbConnectionPool> {
         let pool_builder = DuckDbConnectionPoolBuilder::memory();
-        self.get_or_init_instance_with_builder(Mode::Memory, pool_builder)
-            .await
+        self.get_or_init_instance_with_builder(pool_builder).await
     }
 
     pub async fn get_or_init_file_instance(
@@ -264,24 +263,16 @@ impl DuckDBTableProviderFactory {
         db_path: impl Into<Arc<str>>,
     ) -> Result<DuckDbConnectionPool> {
         let db_path: Arc<str> = db_path.into();
+        let pool_builder = DuckDbConnectionPoolBuilder::file(&db_path);
 
-        let access_mode = match &self.access_mode {
-            AccessMode::ReadOnly => AccessMode::ReadOnly,
-            AccessMode::ReadWrite => AccessMode::ReadWrite,
-            AccessMode::Automatic => AccessMode::Automatic,
-        };
-        let pool_builder =
-            DuckDbConnectionPoolBuilder::file(&db_path).with_access_mode(access_mode);
-
-        self.get_or_init_instance_with_builder(Mode::File, pool_builder)
-            .await
+        self.get_or_init_instance_with_builder(pool_builder).await
     }
 
     pub async fn get_or_init_instance_with_builder(
         &self,
-        mode: Mode,
         pool_builder: DuckDbConnectionPoolBuilder,
     ) -> Result<DuckDbConnectionPool> {
+        let mode = pool_builder.get_mode();
         let key = match mode {
             Mode::File => {
                 let path = pool_builder.get_path();
@@ -289,6 +280,14 @@ impl DuckDBTableProviderFactory {
             }
             Mode::Memory => DbInstanceKey::memory(),
         };
+
+        let access_mode = match &self.access_mode {
+            AccessMode::ReadOnly => AccessMode::ReadOnly,
+            AccessMode::ReadWrite => AccessMode::ReadWrite,
+            AccessMode::Automatic => AccessMode::Automatic,
+        };
+        let pool_builder = pool_builder.with_access_mode(access_mode);
+
         let mut instances = self.instances.lock().await;
 
         if let Some(instance) = instances.get(&key) {
--- a/src/duckdb.rs
+++ b/src/duckdb.rs
@@ -133,11 +133,6 @@ pub enum Error {
         value: String,
         source: byte_unit::ParseError,
     },
-
-    #[snafu(display(
-        r#"Invalid value "{value}" for "connection_pool_size": must be a positive integer greater than zero."#
-    ))]
-    InvalidConnectionPoolSize { value: String },
 }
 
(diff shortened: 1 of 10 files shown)


src/sql/db_connection_pool/duckdbpool.rs
```diff
@@ -15,6 +15,8 @@ use crate::{
     UnsupportedTypeAction,
 };
 
+const DEFAULT_IDLE_CONNECTION: u32 = 10;
```
lukekim, line 18:
```suggestion
const DEFAULT_MIN_IDLE_CONNECTIONS: u32 = 10;
```

src/duckdb.rs
```diff
@@ -279,7 +309,20 @@ impl DuckDBTableProviderFactory {
             return Ok(instance.clone());
         }
 
-        let pool = DuckDbConnectionPool::new_file(&db_path, &self.access_mode)
+        let mut pool_builder = DuckDbConnectionPoolBuilder::default()
+            .with_access_mode(&self.access_mode)
```
peasee, line 313:
```suggestion
            .with_access_mode(Arc::clone(&self.access_mode))
```

**suggestion:** If we can store `AccessMode` in an `Arc<>` on the factory, we can clone the Arc to pass an owned value to `DuckDbConnectionPoolBuilder`.

This will let us remove the lifetime from that struct by doing:

```rust
pub struct DuckDbConnectionPoolBuilder {
...
    access_mode: Arc<AccessMode>,
...
}
```

It will also require updates to `DuckDbConnectionPool::new_file` and `DuckDbConnectionPool::new_memory`.

src/sql/db_connection_pool/duckdbpool.rs
```diff
@@ -54,31 +113,26 @@ impl std::fmt::Debug for DuckDbConnectionPool {
...
-    /// # Returns
-    ///
-    /// * A new `DuckDbConnectionPool`
-    ///
-    /// # Errors
-    ///
-    /// * `DuckDBConnectionSnafu` - If there is an error creating the connection pool
-    /// * `ConnectionPoolSnafu` - If there is an error creating the connection pool
-    pub fn new_memory() -> Result<Self> {
+    fn new_memory_internal(max_size: Option<u32>, min_idle: Option<u32>) -> Result<Self> {
```
peasee, line 121:
```suggestion
    fn new_memory(max_size: Option<u32>, min_idle: Option<u32>) -> Result<Self> {
```

**nit:** I don't think we need to separate this out into an `_internal` function.

The builder can just call `::new_memory(None, None)` itself if it needs to. By introducing a builder, we remove the ability to directly build a pool without using a builder. E.g. getting a basic pool now becomes:

```rust
let pool = DuckDbConnectionPoolBuilder::new().memory().build()?;
```

src/sql/db_connection_pool/duckdbpool.rs
```diff
@@ -31,6 +31,65 @@ pub enum Error {
...
+    /// Create a file-based database connection pool
+    pub fn build_file(self) -> Result<DuckDbConnectionPool> {
+        DuckDbConnectionPool::new_file_internal(
+            self.path.as_str(),
+            self.access_mode,
+            self.max_size,
+            self.min_idle,
+        )
+    }
+}
```
peasee, line 91: **suggestion:** I think a more idiomatic builder pattern for this would be `DuckDbConnectionPoolBuilder::new().memory().build()` or `::new().file().build()` - rather than separate methods to build each.

The builder would default to memory mode, based on the [`Mode` enum](https://github.com/datafusion-contrib/datafusion-table-providers/blob/spiceai/src/sql/db_connection_pool/mod.rs#L37).

```rust
pub fn memory(mut self) -> Self {
self.mode = Mode::Memory;
self
}
```

src/sql/db_connection_pool/duckdbpool.rs
peasee: Ah, I missed this related comment: https://github.com/spiceai/spiceai/pull/5117#discussion_r2011583639

src/duckdb.rs
```diff
@@ -251,14 +251,42 @@ impl DuckDBTableProviderFactory {
...
+
+    pub async fn get_or_init_file_instance(
+        &self,
+        db_path: impl Into<Arc<str>>,
+    ) -> Result<DuckDbConnectionPool> {
+        self.get_or_init_file_instance_with_size_config(db_path, None, None)
+            .await
+    }
+
+    pub async fn get_or_init_memory_instance_with_size_config(
```
phillipleblanc, line 266:
```suggestion
    async fn get_or_init_memory_instance_with_size_config(
```

src/duckdb.rs
peasee: actually, this would result in a change to the existing `::new_file()` signature. Let's avoid making an `Arc<>`, but instead mock-copy the enum (because it doesn't implement `Copy` or `Clone`):

```rust
match &self.access_mode {
            AccessMode::ReadOnly => AccessMode::ReadOnly,
            AccessMode::ReadWrite => AccessMode::ReadWrite,
            AccessMode::Automatic => AccessMode::Automatic,
        }
```

Then we can store a plain `AccessMode` in the builder struct, and pass it like usual to `new_file_internal`.

src/sql/db_connection_pool/duckdbpool.rs
phillipleblanc: Or instead of having a default constructor for the builder, have two separate methods - one each for file or memory, i.e.:

`DuckDbConnectionPoolBuilder::memory()` or `DuckDbConnectionPoolBuilder::file(path)`

src/sql/db_connection_pool/duckdbpool.rs
```diff
@@ -134,6 +188,43 @@ impl DuckDbConnectionPool {
...
+    /// # Returns
+    ///
+    /// * A new `DuckDbConnectionPool`
+    ///
+    /// # Errors
+    ///
+    /// * `DuckDBConnectionSnafu` - If there is an error creating the connection pool
+    /// * `ConnectionPoolSnafu` - If there is an error creating the connection pool
+    pub fn new_memory() -> Result<Self> {
+        Self::new_memory_internal(None, None)
```
phillipleblanc, line 206: Why not just use the builder here instead of having another internal function?

src/sql/db_connection_pool/duckdbpool.rs
```diff
@@ -134,6 +188,43 @@ impl DuckDbConnectionPool {
...
+    /// # Returns
+    ///
+    /// * A new `DuckDbConnectionPool`
+    ///
+    /// # Errors
+    ///
+    /// * `DuckDBConnectionSnafu` - If there is an error creating the connection pool
+    /// * `ConnectionPoolSnafu` - If there is an error creating the connection pool
+    pub fn new_file(path: &str, access_mode: &AccessMode) -> Result<Self> {
+        Self::new_file_internal(path, access_mode, None, None)
```
phillipleblanc, line 225: Same as above - why not just use the builder here?

src/duckdb.rs
phillipleblanc, line 266:
```suggestion
    pub async fn get_or_init_instance_with_builder(
```

This can just take the builder directly - the user constructs the builder they want to use, and then we don't need two separate `get_or_init` functions for memory or file - because its encoded in the builder already.

phillipleblanc: Actually, if we rework it to take the builder directly, then this can be public again, and `get_or_init_file_instance` and `get_or_init_memory_instance` just become convenience methods that initialize the builder and pass to `get_or_init_instance_with_builder`

src/duckdb.rs
```diff
@@ -251,35 +254,49 @@ impl DuckDBTableProviderFactory {
...
+        self.get_or_init_instance_with_builder(Mode::File, pool_builder)
+            .await
     }
 
-    pub async fn get_or_init_file_instance(
+    pub async fn get_or_init_instance_with_builder(
         &self,
-        db_path: impl Into<Arc<str>>,
+        mode: Mode,
+        pool_builder: DuckDbConnectionPoolBuilder,
```
phillipleblanc, line 283: It's unnecessary to pass in the mode when that is already part of the builder

peasee (approved):

Issue #279, opened 2025-03-23
Add Python support for table provider


crystalxyz:
This issue aims to track Python FFI-related development to expose table providers in Python.
It is created as a followup of #264 
Here is a list of todo items:
- [x] Python ODBC table provider + example
- [x] Python Postgres table provider + example
- [x] Python MySQL table provider + example
- [x] Python Flight table provider + example
- [x] Integration testing
- [ ] Finalize Python interface design, i.e. new_memory() or new(":memory:")

phillipleblanc:
@CrystalZhou0529 would you like me to assign you to this issue?

CrystalZhou0529:
@phillipleblanc yes please!

PR #281, opened 2025-03-24
feat: Implement Python ODBC, MySQL, Postgres, Flight table providers

crystalxyz:
Addresses part of #279 
### Summary
This feature implements multiple Python table providers as well as corresponding examples.

### Details
- Implemented ODBC Rust FFI and Python interface
- Implemented MySQL Rust FFI and Python interface
- Implemented Postgres Rust FFI and Python interface
- Implemented Flight Rust FFI and Python interface
- Did minor refactoring of `get_tokio_runtime` to share it among all Python table providers.
- Updated documentation to set up ODBC on ARM Mac and to run examples.

### Test
- Existing Rust tests have all passed
- Verified that all the tests in `/python/examples` work

### Next step
Add integration testing


python/src/odbc.rs
```diff
@@ -1 +1,66 @@
...
+    utils::{to_pyerr, wait_for_future},
+    RawTableProvider,
+};
+
+#[pyclass(module = "datafusion_table_providers._internal.odbc")]
+struct RawODBCTableFactory {
+    _pool: Arc<ODBCPool>,
+    // TODO: 'static lifetime might be wrong, we want the lifetime to be 'py but it is
+    // still unclear how to define it.
+    factory: ODBCTableFactory<'static>,
```
CrystalZhou0529, line 19: Pyclass does not allow lifetime arguments, so I can't really define this struct with `'py`. My temporary solution is to use the weakest `'static` lifetime that lasts for the entire duration of the program. It works but can be a problem due to memory waste.

python/src/odbc.rs
```diff
@@ -1 +1,66 @@
...
+    #[pyo3(signature = (params))]
+    pub fn new(params: &Bound<'_, PyDict>) -> PyResult<Self> {
+        // Convert Python dict into Rust hashmap, and convert it to secret map
+        let mut hashmap = HashMap::new();
+        for (key, value) in params.iter() {
+            let key: String = key.extract()?;
+            let value: String = value.extract()?;
+            hashmap.insert(key, value);
+        }
+        let hashmap = to_secret_map(hashmap);
```
CrystalZhou0529, line 34: This is probably not ideal. In Rust interface, we allow users to construct parameters and call `to_secret_map()` if needed, so users can decide how to construct params. But here in Python interface, I just wrap that into Rust FFI so that users don't need to call `to_secret_map()` in Python.

However, maybe users are always supposed to call `to_secret_map()` so that this implementation is actually fine?  I'm not too sure about the use cases.

CrystalZhou0529:
@timsaucer If you'd like to take a look!

core/src/flight/exec.rs
```diff
@@ -190,7 +191,9 @@ async fn flight_stream(
 ) -> Result<SendableRecordBatchStream> {
     let mut errors: Vec<Box<dyn Error + Send + Sync>> = vec![];
     for loc in partition.locations.iter() {
-        let client = flight_client(loc, grpc_headers.as_ref(), &size_limits).await?;
+        let get_client = || async { flight_client(loc, grpc_headers.as_ref(), &size_limits).await };
+        let client = run_async_with_tokio(get_client).await?;
+        // let client = flight_client(loc, grpc_headers.as_ref(), &size_limits).await?;
```
phillipleblanc, line 196: Why is this required?

core/src/sql/db_connection_pool/runtime.rs
```diff
@@ -0,0 +1,31 @@
+// If calling directly from Rust, there is already tokio runtime so no
+// additional work is needed. If calling from Python FFI, there's no existing
+// tokio runtime, so we need to start a new one.
+use std::{future::Future, sync::OnceLock};
+
+use tokio::runtime::{Handle, Runtime};
+
+pub fn get_tokio_runtime() -> &'static Runtime {
+    // TODO: this function is a repetition of python/src/utils.rs::get_tokio_runtime.
+    // Think about how to refactor it
```
phillipleblanc, line 10:
```suggestion
```

python/python/datafusion_table_providers/odbc.py
```diff
@@ -0,0 +1,35 @@
...
+"""Python interface for ODBC table provider."""
+
+from typing import Any, List
+from . import _internal
+
+class ODBCTableFactory:
+    """ODBC table factory."""
+
+    def __init__(self, params: dict) -> None:
+        """Create am odbc table factory."""
```
phillipleblanc, line 26:
```suggestion
        """Create an odbc table factory."""
```

python/src/odbc.rs
phillipleblanc: The idea with the secret map is that consumers that want to ensure protection of secrets in-memory have the ability to. But given that is a Rust-only interface, doing that conversion here on behalf of Python consumers is fine.

python/src/odbc.rs
```diff
@@ -1 +1,66 @@
...
+        let pool = Arc::new(ODBCPool::new(hashmap).map_err(to_pyerr)?);
+        Ok(Self {
+            factory: ODBCTableFactory::new(pool.clone()),
+            _pool: pool,
+        })
+    }
+
+    pub fn tables(&self) -> PyResult<Vec<String>> {
+        // This method is not supported yet because of unimplemented traits in odbcconn.
+        unimplemented!();
```
phillipleblanc, line 45: Maybe this should return an empty vec instead of panicking?

python/src/odbc.rs
```diff
@@ -1 +1,66 @@
...
+
+    pub fn tables(&self) -> PyResult<Vec<String>> {
+        // This method is not supported yet because of unimplemented traints in odbcconn.
+        unimplemented!();
+    }
+
+    pub fn get_table(&self, py: Python, table_reference: &str) -> PyResult<RawTableProvider> {
+        let table = wait_for_future(
+            py,
+            self.factory.table_provider(table_reference.into(), None),
```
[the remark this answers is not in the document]
phillipleblanc: That is fine - the interface allows consumers to specify the schema if they already know it, otherwise the factory will infer it.

core/src/sql/db_connection_pool/runtime.rs
phillipleblanc: Having it duplicated here is fine.

core/src/sql/db_connection_pool/postgrespool.rs
```diff
@@ -385,7 +385,8 @@ impl
         >,
     > {
         let pool = Arc::clone(&self.pool);
-        let conn = pool.get_owned().await.context(ConnectionPoolRunSnafu)?;
+        let get_conn = async || pool.get_owned().await.context(ConnectionPoolRunSnafu);
+        let conn = run_async_with_tokio(get_conn).await?;
```
phillipleblanc, line 389: It would be nice to have some documentation on when this is required - i.e. as we're making changes to this crate, what do we need to do to ensure this doesn't break the Python providers?

phillipleblanc: It would also be good to have some CI checks for the python table providers to ensure that we don't inadvertently break them going forward - but if you want to handle that as a follow-on PR, I am ok with that too.

phillipleblanc (commented):
Thanks @crystalxyz this is excellent. I've left a few minor comments, but I think this PR is nearly ready to go.

python/src/odbc.rs
crystalxyz: I guess it's fine anyways because I didn't expose `tables()` it in the Python interface, but just leave it here as a reminder for ourselves to add the implementation later.

core/src/sql/db_connection_pool/postgrespool.rs
crystalxyz: > It would be nice to have some documentation on when this is required

I don't think the normal changes would affect this it is needed when we call an async await. I'm doing some experiments on when we need it, but generally it should be safer once we get integration testing added.

phillipleblanc: Thanks, I'm happy to merge this as is and have the documentation come in a follow-up PR as well.

core/src/sql/db_connection_pool/runtime.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+pub async fn run_async_with_tokio<F, Fut, T, E>(f: F) -> Result<T, E>
+where
+    F: FnOnce() -> Fut,
+    Fut: Future<Output = Result<T, E>>,
+{
+    match Handle::try_current() {
+        Ok(_) => f().await,
+        // Err(_) => get_tokio_runtime().0.block_on(f()),
```
phillipleblanc, line 31:
```suggestion
```

phillipleblanc (approved):

phillipleblanc:
@crystalxyz Thank you for this PR! This is epic.

PR #287, opened 2025-03-25
refactor: New TableManager interfaces for DuckDB

peasee:
* Updates the DuckDB `TableManager` to be less coupled to a `DuckDB` instance and removes it storing any connection pool. A `TableManager` is now detached from a connection pool, and most functions require a transaction instance to operate.
* Updates the `TableManager` with a concept of internal vs base tables. Internal tables are prefixed with `__data_` and have a unix timestamp appended to the end with `_{timestamp}` to create unique table names over time. This allows new data inserts to avoid `ALTER TABLE` to rename a table, and instead `CREATE OR REPLACE VIEW` with the name of the table definition.
* Updates to use the `arrow_scan` function to insert data into tables. Wraps the view the table function creates in a `ViewCreator`, so that we can use `ViewCreator::insert_into` to insert the data into our tables with `OnConflict` resolution.
* General refactoring as a result of the above changes
* More unit tests covering additional cases in the writer and covering most functions from the `TableManager`

### DIFF
--- a/src/duckdb/creator.rs
+++ b/src/duckdb/creator.rs
@@ -123,9 +123,17 @@ impl TableDefinition {
             let table_name = row
                 .get::<usize, String>(0)
                 .context(super::UnableToQueryDataSnafu)?;
-            let Some(timestamp) = table_name.split('_').last() else {
+            // __data_{table_name}% could be a subset of another table name, so we need to check if the table name starts with the table definition name
+            let inner_name = table_name.replace("__data_", "");
+            let mut parts = inner_name.split('_');
+            let Some(timestamp) = parts.next_back() else {
                 continue; // skip invalid table names
             };
+
+            let inner_name = parts.join("_");
+            if inner_name != self.name.to_string() {
+                continue;
+            }
 
             let timestamp = timestamp
                 .parse::<u64>()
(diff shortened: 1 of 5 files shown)


src/duckdb/creator.rs
```diff
@@ -334,69 +364,110 @@ impl TableCreator {
     }
 
     #[tracing::instrument(level = "debug", skip_all)]
-    fn drop_table(&self, transaction: &Transaction<'_>) -> super::Result<()> {
-        let sql = format!(r#"DROP TABLE IF EXISTS "{}""#, self.table_name);
-        tracing::debug!("{sql}");
+    pub(crate) fn list_internal_tables(
```
phillipleblanc, line 367: Is this function really, list internal tables that aren't this one?

src/duckdb/creator.rs
```diff
@@ -1,324 +1,389 @@
...
 
-        Ok(duckdb)
+    /// Returns the canonical name for this table, which is the internal name if the table is internal, or the table name if it is not.
+    pub(crate) fn table_name(&self) -> &RelationName {
+        if self.is_internal {
+            &self.internal_name
+        } else {
+            &self.table_definition.name
+        }
     }
```
phillipleblanc, line 182:
```suggestion
    pub(crate) fn table_name(&self) -> &RelationName {
        if let Some(internal_name) = self.internal_name {
            internal_name&
        } else {
            &self.table_definition.name
        }
    }
```

src/duckdb/creator.rs
```diff
@@ -1,324 +1,388 @@
...
-    ) -> super::Result<HashSet<String>> {
-        // DuckDB provides convinient querable 'pragma_table_info' table function
-        // Complex table name with schema as part of the name must be quoted as
-        // '"<name>"', otherwise it will be parsed to schema and table name
-        let sql = format!(
-            r#"SELECT name FROM pragma_table_info('{table_name}') WHERE pk = true"#,
-            table_name = quote_identifier(table_name)
-        );
-        tracing::debug!("{sql}");
+    /// Creates the table for this `TableCreator`. Does not create indexes - use `TableCreator::create_indexes` to apply indexes.
```
phillipleblanc, line 214:
```suggestion
    /// Creates the table for this `TableManager`. Does not create indexes - use `TableManager::create_indexes` to apply indexes.
```

src/duckdb/creator.rs
```diff
@@ -333,68 +397,288 @@ impl TableCreator {
         Ok(create_stmt)
     }
 
+    /// List all internal tables related to this table creators table definition.
```
phillipleblanc, line 400:
```suggestion
    /// List all internal tables related to this table manager's table definition.
```

phillipleblanc (approved):

src/duckdb/creator.rs
```diff
@@ -1,324 +1,388 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the internal tables cannot be listed.
+    pub fn list_internal_tables(
+        &self,
+        tx: &Transaction<'_>,
+    ) -> super::Result<Vec<(RelationName, u64)>> {
+        // list all related internal tables, based on the table definition name
+        let sql = format!(
```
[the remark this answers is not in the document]
peasee: For a `TableDefinition`, this is intended. If you call `TableManager::list_other_internal_tables()`, it will exclude itself. However, if you're listing internal tables from a definition you likely do not have a reference to a `TableManager`

ewgenius (approved):

src/duckdb/creator.rs
```diff
@@ -1,324 +1,388 @@
...
+    ) -> super::Result<()> {
+        let table_name = self.table_name();
+
+        let unique = index.1 == IndexType::Unique;
+        let columns = index.0;
+        let mut index_builder = IndexBuilder::new(&table_name.to_string(), columns);
+        if unique {
+            index_builder = index_builder.unique();
+        }
+        let sql = index_builder.build_postgres();
```
Sevenannn, line 315: Why is the index_builder using `build_postgres` here? Looks like the query is directly sent to DuckDB without going through spice and need a duckdb variant syntax

Sevenannn: Also the build_postgres is tied to Postgres feature and it'll be unavailable when only building with DuckDB

phillipleblanc: > Also the build_postgres is tied to Postgres feature and it'll be unavailable when only building with DuckDB

This isn't true

Sevenannn (approved):

sgrebnov (approved):

PR #288, opened 2025-03-25
Switch from Appender to duckdb_arrow_scan

phillipleblanc:
Switches from using the Appender API to insert Arrow batches into DuckDB to using the `duckdb_arrow_scan` API.

This gives us several benefits:
- The `duckdb_arrow_scan` API is more efficient - it leverages zero-copy semantics for Arrow batches whereas the current VTab implementation for Arrow in the duckdb-rs crate will copy data when converting from Arrow to the internal DuckDB format.
- It allows for streaming of the record batches - which allows for processing data that is larger than memory.

### DIFF
--- a/src/duckdb/write.rs
+++ b/src/duckdb/write.rs
@@ -1,3 +1,4 @@
+use std::time::{SystemTime, UNIX_EPOCH};
 use std::{any::Any, fmt, sync::Arc};
 
 use crate::duckdb::DuckDB;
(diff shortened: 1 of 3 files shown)


lukekim (approved):

sgrebnov (approved):

src/duckdb/write.rs
```diff
@@ -392,31 +398,64 @@ impl DisplayAs for DuckDBDataSink {
...
+        schema,
+    )));
+
+    let current_ts = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .context(super::UnableToGetSystemTimeSnafu)
+        .map_err(to_datafusion_error)?
+        .as_millis();
+
+    let view_name = format!("__scan_{}_{current_ts}", table.table_name());
```
peasee, line 415:
```suggestion
    let view_name = format!("__scan_{}_{current_ts}", table.table_name());
```

`table.table_name()` will be the internal table name with a unix timestamp, if it is an internal table. Might want `table.definition_name()`, for the name from the table definition only?

phillipleblanc: It shouldn't matter - but I'd like to keep it as the internal name, since it represents what we are ingesting into.

PR #289, opened 2025-03-26
Support setting `temp_directory` in DuckDB if present in the CreateExternalTable options.

phillipleblanc:
Sets the DuckDB parameter `temp_directory` if present in the CreateExternalTable options.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,7 @@ rstest = "0.22.0"
 geozero = { version = "0.13.0", features = ["with-wkb"] }
 tokio-stream = { version = "0.1.15", features = ["net"] }
 insta = { version = "1.40.0", features = ["filters"] }
+datafusion-physical-plan = { version = "43" }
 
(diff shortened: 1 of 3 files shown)


ewgenius (approved):

Cargo.toml
```diff
@@ -78,6 +78,7 @@ rstest = "0.22.0"
 geozero = { version = "0.13.0", features = ["with-wkb"] }
 tokio-stream = { version = "0.1.15", features = ["net"] }
 insta = { version = "1.40.0", features = ["filters"] }
+datafusion-physical-plan = { version = "43" }
```
peasee, line 81: is this necessary?

phillipleblanc: Yes, otherwise the tests for Duckdb fail when run by themselves. This is only dev dependencies

peasee (approved):

sgrebnov (approved):

PR #320, opened 2025-04-21
feat: Create Python integration testing

crystalxyz:
Addresses part of #279 

## Summary
Add integration testing for Python interfaces and update github workflow to run Python tests.

Implemented tests for SQLite, DuckDB, ODBC SQLite, MySQL and PostgresSQL.

## Testing
Verified locally, pipeline testing passed.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -6,6 +6,8 @@ on:
     branches:
       - main
       - spiceai
+      # TODO: remove this once we have fully tested the python integration tests
+      - python-integration-test
 
 jobs:
   lint:
@@ -107,3 +109,10 @@ jobs:
 
       - name: Run tests
         run: make test
+
+      - name: Run Python tests
+        run: |
+          cd python/python
+          maturin develop
+          cd tests
+          python -m pytest tests/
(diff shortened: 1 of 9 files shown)


.github/workflows/pr.yaml
```diff
@@ -6,6 +6,8 @@ on:
     branches:
       - main
       - spiceai
+      # TODO: remove this once we have fully tested the python integration tests
+      - python-integration-test
```
crystalxyz, line 10: This needs to be removed after testing is done.

.github/workflows/pr.yaml
```diff
@@ -107,3 +113,16 @@ jobs:
 
       - name: Run tests
         run: make test
+
+      - name: Build Python package
```
phillipleblanc, line 117: Can we run this in a separate job? The test workflow already takes a long time to run.

crystalxyz: Make sense to me, will do it

lukekim:
Nice work!

.github/workflows/pr.yaml
phillipleblanc: I think this PR is good to go once this is removed.

crystalxyz: Just removed it. Once the pipeline passes we should be good to go

phillipleblanc (approved):

phillipleblanc:
@crystalxyz Thanks!

PR #334, opened 2025-05-10
MySQL: Set the `character_set_results`/`character_set_client`/`character_set_connection` session variables on connection setup

Sevenannn:
Sets the `character_set_results`/`character_set_client`/`character_set_connection` session variables on connection setup. Also moves the `SET time_zone = '+00:00'` query into the same setup function.

See MySQL character set results doc for an explanation on these variables: https://dev.mysql.com/doc/refman/8.4/en/charset-connection.html

This is required to ensure that the results we receive from MySQL are encoded correctly into UTF8

### DIFF
--- a/src/sql/db_connection_pool/mysqlpool.rs
+++ b/src/sql/db_connection_pool/mysqlpool.rs
@@ -49,6 +49,9 @@ pub enum Error {
 
     #[snafu(display("{message}\nEnsure the given MySQL database exists"))]
     UnknownMySQLDatabase { message: String },
+
+    #[snafu(display("Unsupported value: '{value}' for 'character_set_results'. Only 'utf8mb3' and 'utf8mb4' are supported."))]
+    InvalidCharacterSetResults { value: String },
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


Jeadie (approved):

src/sql/db_connection_pool/mysqlpool.rs
```diff
@@ -160,6 +163,24 @@ impl MySQLConnectionPool {
 
         connection_string = connection_string.ssl_opts(ssl_opts);
 
+        if let Some(mysql_character_set_results) = params
+            .get("character_set_results")
+            .map(SecretBox::expose_secret)
+        {
+            match mysql_character_set_results {
+                "utf8mb3" | "utf8mb4" => {
```
phillipleblanc, line 171: I don't think this should be an option, we should always set it to `utf8mb4`

phillipleblanc (approved):

src/sql/db_connection_pool/mysqlpool.rs
phillipleblanc: Let's always set this to utf8mb4

phillipleblanc (changes requested):
We shouldn't make this a configuraiton option, we should always set it to utf8mb4

sgrebnov (approved):

phillipleblanc (approved):

ewgenius (approved):

Issue #345, opened 2025-05-15
SQL execute error: SELECT count(*) FROM postgres.public.companies


mmooyyii:
run the postgres.rs in example;
```
// Query Example 1: Query the renamed table through default catalog
    let df = ctx
        .sql("SELECT count(*) FROM postgres.public.companies")
        .await
        .expect("select failed");
    df.show().await.expect("show failed");
```

```
/Users/yimo/.cargo/bin/cargo run --color=always --example postgres --profile dev --manifest-path /Users/yimo/PycharmProjects/datafusion-table-providers/core/Cargo.toml
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.29s
     Running `target/debug/examples/postgres`

thread 'main' panicked at core/examples/postgres.rs:106:21:
show failed: Internal("Physical input schema should be the same as the one converted from logical input schema. Differences: \n\t- Different number of fields: (physical) 1 vs (logical) 0")
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/d7ea436a02d5de4033fcf7fd4eb8ed965d0f574c/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/d7ea436a02d5de4033fcf7fd4eb8ed965d0f574c/library/core/src/panicking.rs:75:14
   2: core::result::unwrap_failed
             at /rustc/d7ea436a02d5de4033fcf7fd4eb8ed965d0f574c/library/core/src/result.rs:1704:5
   3: core::result::Result<T,E>::expect
             at /Users/yimo/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:1061:23
   4: postgres::main::{{closure}}
             at ./core/examples/postgres.rs:106:5
   5: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /Users/yimo/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/future/future.rs:124:9
   6: tokio::runtime::park::CachedParkThread::block_on::{{closure}}
             at /Users/yimo/.cargo/registry/src/mirrors.ustc.edu.cn-5857e57f01837ef8/tokio-1.45.0/src/runtime/park.rs:284:60
   7: tokio::task::coop::with_budget
[...]

PR #347, opened 2025-05-16
fix SQL: select count(*) from table

mmooyyii:
fix https://github.com/datafusion-contrib/datafusion-table-providers/issues/345
I don't know how to add unit test, so I add it to example.

### DIFF
--- a/core/src/sql/sql_provider_datafusion/mod.rs
+++ b/core/src/sql/sql_provider_datafusion/mod.rs
@@ -229,8 +229,6 @@ impl<T, P> Display for SqlTable<T, P> {
     }
 }
 
-static ONE_COLUMN_SCHEMA: LazyLock<SchemaRef> =
-    LazyLock::new(|| Arc::new(Schema::new(vec![Field::new("1", DataType::Int64, true)])));
 
 pub fn project_schema_safe(
     schema: &SchemaRef,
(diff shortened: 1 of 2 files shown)


core/src/sql/sql_provider_datafusion/mod.rs
```diff
@@ -229,23 +229,14 @@ impl<T, P> Display for SqlTable<T, P> {
...
-static ONE_COLUMN_SCHEMA: LazyLock<SchemaRef> =
-    LazyLock::new(|| Arc::new(Schema::new(vec![Field::new("1", DataType::Int64, true)])));
 
 pub fn project_schema_safe(
     schema: &SchemaRef,
     projection: Option<&Vec<usize>>,
 ) -> DataFusionResult<SchemaRef> {
     let schema = match projection {
         Some(columns) => {
-            if columns.is_empty() {
```
phillipleblanc, line 241: Hmm, the reason this was added is because its needed for the unparser, so this change will break the federated table case.

mmooyyii: > Hmm, the reason this was added is because its needed for the unparser, so this change will break the federated table case.

for example? I try fix them.

phillipleblanc: It is the same SQL, but its when its used with the `datafusion-federation` crate.

PR #356, opened 2025-05-24
Support custom scalar functions in MySQL unparser dialect

phillipleblanc:
Allows the MySQLTableFactory to have an optional customized MySQL dialect.

Adds a new MySQLDialect that implements `with_custom_scalar_overrides` to allow consumers to specify their own MySQL scalar function overrides.

### DIFF
--- a/src/mysql/dialect.rs
+++ b/src/mysql/dialect.rs
@@ -1,0 +1,95 @@
+use std::sync::Arc;
+
+use arrow_schema::TimeUnit;
+use datafusion::{
+    common::HashMap,
+    sql::unparser::{
+        dialect::{
+            DateFieldExtractStyle, Dialect, IntervalStyle, MySqlDialect as DFMySqlDialect,
+            ScalarFnToSqlHandler,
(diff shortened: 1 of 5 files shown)


src/mysql/dialect.rs
```diff
@@ -0,0 +1,98 @@
...
+            dialect::{
+                DateFieldExtractStyle, Dialect, IntervalStyle, MySqlDialect as DFMySqlDialect,
+                ScalarFnToSqlHandler,
+            },
+            Unparser,
+        },
+    },
+};
+
+pub struct MySqlDialect {
```
sgrebnov, line 19: Should we move this into Datafusion crate to be consistent with DuckDB dialect?
https://github.com/apache/datafusion/blob/a256eb4e45d85701e9535c32eb84260c6817522a/datafusion/sql/src/unparser/dialect.rs#L365

sgrebnov (approved):

phillipleblanc:
Turns out I don't need this change

PR #369, opened 2025-06-18
Clickhouse: Add read only table provider for clickhouse

trueleo:
## Description:
This PR add ClickHouse table provider implementation. 

## Summary:
- Adds ClickHouseConnectionPool and ClickHouseConnection which uses official clickhouse client to query data. Fetches data using [ArrowStream](https://clickhouse.com/docs/interfaces/formats/ArrowStream) format. 
- ClickHouseTableFactory currently creates read only table provider for clickhouse table and view ( including parameterized views )

## What's not in here 
- Ability to create and insert data to ClickHouse
- Python binding

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -462,6 +462,7 @@ checksum = "73a47aa0c771b5381de2b7f16998d351a6f4eb839f1e13d48353e17e873d969b"
 dependencies = [
  "bitflags 2.9.1",
  "serde",
+ "serde_json",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


phillipleblanc (approved):

phillipleblanc:
@trueleo Thanks! When trying to submit this to crates, I get this error:

Caused by:
  the remote server responded with an error (status 400 Bad Request): wildcard (`*`) dependency constraints are not allowed on crates.io. Crate with this problem: `clickhouse` See https://doc.rust-lang.org/cargo/faq.html#can-libraries-use--as-a-version-for-their-dependencies for more information
  
Can you raise another PR to constrain the clickhouse crate?

trueleo:
@phillipleblanc Yeah sorry about that :p. Will raise another PR

PR #386, opened 2025-07-11
Add readonly table provider for MongoDB

krinart:
What this PR includes:
* `MongoDBTableFactory`
* Tests (unit + integration)
* Example (under `core/examples`)
* Documentation in `README`
* Unsupported types will be converted to string

Example:
```
    let mongodb_params = to_secret_map(HashMap::from([
        (
            "connection_string".to_string(),
            "mongodb://root:password@localhost:27017/mongo_db?authSource=admin".to_string(),
        ),
        ("sslmode".to_string(), "disabled".to_string()),
    ]));

    let mongodb_pool = Arc::new(
        MongoDBConnectionPool::new(mongodb_params)
            .await
            .expect("unable to create MongoDB connection pool"),
    );

    let table_factory = MongoDBTableFactory::new(mongodb_pool.clone());

    let ctx = SessionContext::new();
    ctx.register_table(
        "companies_v2",
        table_factory
            .table_provider(TableReference::bare("companies"))
            .await
            .expect("failed to register table provider"),
    )
    .expect("failed to register table");

    let df = ctx
        .sql("SELECT * FROM datafusion.public.companies_v2")
        .await
        .expect("select failed");
    df.show().await.expect("show failed");
```

Will output:
```
+--------------------------+----+------------------+
| _id                      | id | name             |
+--------------------------+----+------------------+
| 687077cde4c00398f4f47070 | 1  | Acme Corporation |
| 687077d8694684245cf47070 | 2  | Google           |
+--------------------------+----+------------------+
```

The code is added under `src/mongodb` module. I decided to avoid putting anything under `src/sql` since there's no sql-related logic.

What this PR does not include (next steps):
* No `MongoDBTableProviderFactory`
* No federation support
* No python support

Here's the implemented BSON=>Arrow mapping
```
        Bson::Double => DataType::Float64,
        Bson::String => DataType::Utf8,
        Bson::Array => DataType::List,               Arrays are represented as lists of strings
        Bson::Document => DataType::Utf8,     Nested documents are converted to a string
        Bson::Boolean => DataType::Boolean,
        Bson::Null => DataType::Null,
        Bson::RegularExpression => DataType::Utf8,
        Bson::JavaScriptCodex => DataType::Utf8,
        Bson::JavaScriptCodeWithScope => DataType::Utf8,
        Bson::Int32 => DataType::Int32,
        Bson::Int64 => DataType::Int64,
        Bson::Timestamp => DataType::Timestamp,
        Bson::Binary => DataType::Binary,
        Bson::ObjectId => DataType::Utf8, 
        Bson::DateTime => DataType::Timestamp,
        Bson::Symbol => DataType::Utf8,
        Bson::Decimal128 => DataType::Decimal128,
        Bson::Undefined => DataType::Null,
        Bson::MaxKey => DataType::Utf8,
        Bson::MinKey => DataType::Utf8,
        Bson::DbPointer => DataType::Utf8,
```


tests/postgres/mod.rs
```diff
@@ -285,39 +287,49 @@ async fn test_postgres_jsonb_type(port: usize) {
...
-            .to_string(),
-        serde_json::from_str::<Value>("null").unwrap().to_string(),
-        serde_json::from_str::<Value>(r#"{"nested":{"key":"value"}}"#)
-            .unwrap()
-            .to_string(),
+        r#"{"name": "John", "age": 30}"#,
+        r#"{"name": "Jane", "age": 25}"#,
+        "[1, 2, 3]",
+        "null", 
+        r#"{"nested": {"key": "value"}}"#,
```
krinart, line 295: This test was failing due to order of order of the fields in the string. I refactored it to compare json values directly, not strings.

Not sure why it happened. Maybe some library version got updated.

tests/docker/mod.rs
```diff
@@ -134,7 +134,7 @@ impl<'a> ContainerRunner<'a> {
                 format!("{container_port}/tcp"),
                 Some(vec![PortBinding {
                     host_ip: Some("127.0.0.1".to_string()),
-                    host_port: Some(format!("{host_port}/tcp")),
+                    host_port: Some(format!("{host_port}")),
                 }]),
```
krinart, line 138: Without this fix containers' port was not properly configured, which made containers inaccessible

src/mongodb/table.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    fn schema(&self) -> SchemaRef {
+        Arc::clone(&self.schema)
+    }
+
+    fn table_type(&self) -> TableType {
+        TableType::Base
+    }
+
+    async fn scan(
```
phillipleblanc, line 67: You will need to implement `supports_filters_pushdown` or else DataFusion will not push down any filters, and you will end up doing a full collection scan on every query. I assume this is a bug since you implement support for converting filters to MongoDB Documents.

tests/mongodb/mod.rs
```diff
@@ -0,0 +1,707 @@
...
+        .expect("MongoDB container to start");
+
+    tracing::debug!("MongoDB Container started");
+
+    running_container
+}
+
+#[rstest]
+#[test_log::test(tokio::test)]
+async fn test_mongodb_arrow_oneway() {
```
phillipleblanc, line 692: This test is failing for me locally with:

```
thread 'mongodb::test_mongodb_arrow_oneway' panicked at tests/mongodb/mod.rs:658:5:
assertion `left == right` failed
  left: RecordBatch { schema: Schema { fields: [Field { name: "int32_field", data_type: Int32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "int64_field", data_type: Int64, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "double_field", data_type: Float64, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "decimal_field", data_type: Decimal128(18, 6), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }], metadata: {} }, columns: [PrimitiveArray<Int32>
[
  2147483647,
], PrimitiveArray<Int64>
[
  9223372036854775807,
], PrimitiveArray<Float64>
[
  3.14159265359,
], PrimitiveArray<Decimal128(18, 6)>
[
  123456000,
]], row_count: 1 }
 right: RecordBatch { schema: Schema { fields: [Field { name: "int32_field", data_type: Int32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "int64_field", data_type: Int64, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "double_field", data_type: Float64, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "decimal_field", data_type: Decimal128(38, 10), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }], metadata: {} }, columns: [PrimitiveArray<Int32>
[
  2147483647,
], PrimitiveArray<Int64>
[
  9223372036854775807,
], PrimitiveArray<Float64>
[
  3.14159265359,
], PrimitiveArray<Decimal128(38, 10)>
[
  1234560000000,
]], row_count: 1 }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test mongodb::test_mongodb_arrow_oneway ... FAILED
```

phillipleblanc (commented):
This is a really well-structured PR! Thank you.

I have a couple of comments, but I think this is really close.

src/mongodb/table.rs
krinart: Great catch, thanks!

tests/mongodb/mod.rs
krinart: Great catch! it turns out I messed up with tests, and was running `flight` tests instead of mongo - that's why this issue was not caught by CI.

src/mongodb/table.rs
```diff
@@ -0,0 +1,234 @@
...
+            filters,
+            limit,
+        )?))
+    }
+
+    fn supports_filters_pushdown(
+        &self,
+        filters: &[&Expr],
+    ) -> DataFusionResult<Vec<TableProviderFilterPushDown>> {
+        Ok(vec![TableProviderFilterPushDown::Exact; filters.len()])
```
phillipleblanc, line 82: Instead of claiming we support all filters, we should only claim to support filters that can be parsed by `expr_to_mongo_filter` - that will still allow DataFusion to push as much down as it can, but then it can handle any other filters that mongo can't - instead of erroring like the current implementation would do.

PR #397, opened 2025-07-17
Quote column names in DuckDB queries

kczimm:
Column names that use keywords can no longer be parsed in DuckDB without quoting them. 

This PR adds a failing test and fix.


phillipleblanc:
I will revert the change that caused this bug, so this isn't needed.

PR #406, opened 2025-07-25
Add AttachedDuckDBTableProviderFactory for ATTACHed database files

kczimm:


src/sql/db_connection_pool/duckdbpool.rs
```diff
@@ -314,6 +317,14 @@ impl DuckDbConnectionPool {
             ))))
         }
     }
+
+    pub fn attach_file(&self, file_path: &str) -> Result<()> {
```
mach-kernel, line 321: There's a `get_attachments` function in `DuckDbConnectionPool`, is there any value in (for mode `Memory`) running a query to get the attachments?

```
D select * from duckdb_databases();
┌───────────────┬──────────────┬────────────────┬─────────┬───────────────────────────┬──────────┬─────────┬──────────┐
│ database_name │ database_oid │      path      │ comment │           tags            │ internal │  type   │ readonly │
│    varchar    │    int64     │    varchar     │ varchar │   map(varchar, varchar)   │ boolean  │ varchar │ boolean  │
├───────────────┼──────────────┼────────────────┼─────────┼───────────────────────────┼──────────┼─────────┼──────────┤
│ memory        │          570 │ NULL           │ NULL    │ {}                        │ false    │ duckdb  │ false    │
│ testdb0       │         1974 │ testdb0.duckdb │ NULL    │ {storage_version=v1.0.0+} │ false    │ duckdb  │ false    │
```

There's also the `attached_databases` Vec on that struct that gets populated on pool connect:

https://github.com/datafusion-contrib/datafusion-table-providers/blob/e56af46ed124fe6f597149d2d05a8f2cf0dacc3e/core/src/sql/db_connection_pool/duckdbpool.rs#L332-L344

src/duckdb/write.rs
```diff
@@ -591,18 +590,21 @@ fn write_to_table(
...
     )));
 
-    let current_ts = SystemTime::now()
-        .duration_since(UNIX_EPOCH)
-        .context(super::UnableToGetSystemTimeSnafu)
-        .map_err(to_datafusion_error)?
-        .as_millis();
+    let view_name = table
+        .table_name()
+        .generate_internal_name("scan")
```
peasee, line 595: this results in `__data_scan`. `__data` is used which implies the relation is an internal data table. If someone creates a table named `scan`, this would introduce a bug, as the table manager scans for relations prefixed with `__data_{table_name}`.

Although this is a view, I would prefer we don't re-use this to avoid introducing this bug.

src/sql/db_connection_pool/duckdbpool.rs
```diff
@@ -314,6 +317,14 @@ impl DuckDbConnectionPool {
...
+        tracing::debug!("attaching {file_path}");
+        let conn = self.pool.get()?;
+        conn.execute(&format!("ATTACH '{file_path}'"), [])
```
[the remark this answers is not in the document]
sgrebnov: We previously saw issues with WAL/CHECKPOINTing, the only reliable approach we found is to ensure all connections (pool) are closed which forced CHECKPOINT/WAL. Manually calling was not very reliable and there are cases where it was failing when executing..

src/duckdb/write.rs
peasee: ah, I missed this diff. I had checked out the wrong branch :facepalm:

PR #410, opened 2025-08-01
Update to DF 48

mach-kernel:
## 🗣 Description
- Updates to DF 48, only change is Expr::Literal node new map param


adriangb:
DataFusion 49 is out now. Should we update to that instead?

adriangb:
Yep I found the other PR

phillipleblanc (approved):

PR #420, opened 2025-08-19
fix: Passthrough constaints from TableProviderFactory

peasee:
* Passes through any constraints specified in the `cmd.constraints` for a `TableProviderFactory` into the underlying `SqlTable` base table.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,7 +84,6 @@ datafusion-physical-plan = { version = "48.0.1" }
 tempfile = "3.8.1"
 
 [features]
-default = ["duckdb", "mysql", "postgres", "sqlite"]
 mysql = ["dep:mysql_async", "dep:async-stream"]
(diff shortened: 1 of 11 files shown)


src/sql/sql_provider_datafusion/mod.rs
```diff
@@ -122,9 +125,20 @@ impl<T, P> SqlTable<T, P> {
             table_reference: table_reference.into(),
             engine,
             dialect: None,
+            constraints: None,
         }
     }
 
+    pub fn set_constraints(mut self, constraints: Option<Constraints>) -> Self {
```
phillipleblanc, line 132: the convention we've used elsewhere is to call this `with_constraints_opt`

Cargo.toml
```diff
@@ -84,6 +84,7 @@ datafusion-physical-plan = { version = "48.0.1" }
 tempfile = "3.8.1"
 
 [features]
+default = ["duckdb", "mysql", "postgres", "sqlite"]
```
phillipleblanc, line 87: Why this change?

peasee: I thought I reverted this. My `rust-analyzer` isn't building the project correctly even with `"rust-analyzer.check.features": "all",`

phillipleblanc (approved):

PR #423, opened 2025-08-24
Add controls for upsert behavior to modify batches to remove duplicates and/or pick the latest row on constraint violations that occur within a batch

phillipleblanc:
Adds two new behaviors for the `OnConflict::Upsert` behavior that affects conflicts that occur within a batch. The previous behavior was to error out, as this matches the behavior of most database systems. However, we can be smarter and apply some mitigations to the data of the incoming batch to allow these updates to proceed. These mitigations are opt-in, as they do introduce some overhead compared to not using them.

- `remove_duplicates`: Will run the equivalent of a `SELECT DISTINCT * FROM <batch>` if a violation is detected in the batch.
- `last_write_wins`: When duplicates are found in the batch, keep the row with the highest row number (i.e. the equivalent behavior if upsert were to happen one row at a time)

These options are specified using the existing `on_conflict` option in the `TableProviderFactory::create` method and are given by suffixing the `upsert:{col}` with a `#` and then specifying the options (comma separated). i.e.:

Examples:
`on_conflict: upsert:col1#remove_duplicates` -> Specifies that when a constraint violation occurs on `col1` in the incoming batch, attempt to remove duplicates.
`on_conflict: upsert:(col1,col2)#remove_duplicates,last_write_wins` -> When a constraint violation occurs on the compound primary key `(col1, col2)` then attempt to remove duplicates, and also keep only the last row.

### DIFF
--- a/src/util/constraints.rs
+++ b/src/util/constraints.rs
@@ -54,6 +54,10 @@ impl ValidationOptions {
             ..self
         }
     }
+
+    pub fn is_default(&self) -> bool {
+        !self.remove_duplicates && !self.last_write_wins
+    }
 }
 
 /// The goal for this function is to determine if all of the data described in `batches` conforms to the constraints described in `constraints`.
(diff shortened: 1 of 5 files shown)


src/util/constraints.rs
```diff
@@ -90,7 +236,87 @@ async fn validate_batch_with_constraint(
...
+
+    for batch in batches {
+        let num_rows = batch.num_rows();
+        let row_numbers: Vec<i64> =
+            (global_row_number..global_row_number + num_rows as i64).collect();
+        global_row_number += num_rows as i64;
+
+        // Create new schema with additional row_num column
+        let mut fields = batch.schema().fields().iter().cloned().collect::<Vec<_>>();
+        fields.push(Arc::new(Field::new("__row_num", DataType::Int64, false)));
```
peasee, line 257:
```suggestion
        fields.push(Arc::new(Field::new(BATCH_ROW_NUMBER_COLUMN_NAME, DataType::Int64, false)));
```

nit: magic column name

src/util/constraints.rs
```diff
@@ -90,7 +236,87 @@ async fn validate_batch_with_constraint(
...
+    Ok(batches_with_row_nums)
+}
+
+/// Apply last write wins logic using window functions to keep only the row with highest row number for each unique key
+async fn apply_last_write_wins(
+    df: &mut datafusion::dataframe::DataFrame,
+    unique_fields: &[&arrow::datatypes::Field],
+) -> datafusion::error::Result<datafusion::dataframe::DataFrame> {
+    use datafusion::functions_aggregate::min_max::max_udaf;
+    use datafusion::logical_expr::{WindowFrame, WindowFunctionDefinition};
```
peasee, line 280: nit: use at module, not within function

src/util/constraints.rs
```diff
@@ -90,7 +236,87 @@ async fn validate_batch_with_constraint(
...
+    let partition_by: Vec<Expr> = unique_fields
+        .iter()
+        .map(|field| col(field.name()))
+        .collect();
+
+    // Create a MAX window function to get the maximum row number for each partition
+    let max_row_num_expr = Expr::WindowFunction(datafusion::logical_expr::expr::WindowFunction {
+        fun: WindowFunctionDefinition::AggregateUDF(max_udaf()),
+        params: WindowFunctionParams {
+            args: vec![col("__row_num")],
```
peasee, line 292: nit: magic column name

src/util/constraints.rs
```diff
@@ -90,7 +236,87 @@ async fn validate_batch_with_constraint(
...
+            partition_by,
+            order_by: vec![col("__row_num").sort(false, true)], // Order by row_num DESC
```
peasee, line 294: nit: magic column name

src/util/constraints.rs
```diff
@@ -90,7 +236,87 @@ async fn validate_batch_with_constraint(
...
+            window_frame: WindowFrame::new(Some(false)),
+            null_treatment: None,
+        },
+    });
+
+    // Add the maximum row number as a column with explicit alias
+    let df_with_max = df
+        .clone()
+        .with_column("max_row_num_col", max_row_num_expr)?;
```
peasee, line 303: nit: magic alias name

PR #431, opened 2025-08-27
Use dialect-specific quote style for projection in 'SqlExec'

Jeadie:
Users can specify a dialect in `SqlTable` that will get ignored in `SqlExec` when constructing the appropriate SQL query. This PR does two things:
 1. Provides the Dialect from `SqlTable` into `SqlExec`.
 2. Uses the `Dialect` to determine how to quote identifiers.

Currently `SQLTable` does not work for custom Dialects where `Dialect::identifier_quote_style() != '"'` (e.g. Databricks), even when the appropriate dialect is provided to `SqlTable`, e.g.
```rust
SqlTable::new_with_schema("databricks", &self.pool, schema, table_reference, None)
    .with_dialect(dialect::CustomDialectBuilder::new()
        .with_identifier_quote_style('`')
        .with_interval_style(dialect::IntervalStyle::MySQL)
        .build().into())
```

### DIFF
--- a/src/sql/sql_provider_datafusion/mod.rs
+++ b/src/sql/sql_provider_datafusion/mod.rs
@@ -307,16 +307,23 @@ impl<T, P> SqlExec<T, P> {
     }
 
     pub fn sql(&self) -> Result<String> {
-        let dialect = self.dialect.clone().unwrap_or(self.engine.map_or(
-            Arc::new(DefaultDialect {}) as Arc<dyn Dialect + Send + Sync>,
-            |e| e.dialect(),
-        ));
         let columns = self
             .projected_schema
             .fields()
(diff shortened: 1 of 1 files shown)


Jeadie:
Ideally we refactor `SqlExec` to be more basic, for example
```rust
#[derive(Clone)]
pub struct SqlExec<T, P> {
    projected_schema: SchemaRef,
    pool: Arc<dyn DbConnectionPool<T, P> + Send + Sync>,
    properties: PlanProperties,
    sql: String,
}
```

But it is currently used as a`ExecutionPlan` that will construct SQL for other uses cases (e.g. for `DuckSqlExec`, `MySQLSQLExec`, etc).

PR #474, opened 2025-10-25
DuckDB: include ANALYZE statement after write to update query optimizer statistics

sgrebnov:
This helps DuckDB's query planner generate better execution plans, especially after large data changes.

https://duckdb.org/docs/stable/sql/statements/analyze

### DIFF
--- a/core/src/duckdb.rs
+++ b/core/src/duckdb.rs
@@ -148,6 +148,12 @@ pub enum Error {
     #[snafu(display("Failed to parse the system time: {source}"))]
     UnableToParseSystemTime { source: std::num::ParseIntError },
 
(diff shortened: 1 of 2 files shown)


core/src/duckdb/write.rs
```diff
@@ -707,6 +711,30 @@ fn write_to_table(
     Ok(rows as u64)
 }
 
+/// Executes an ANALYZE statement to update query optimizer statistics.
+/// This helps DuckDB's query planner generate better execution plans, especially after large data changes.
+/// https://duckdb.org/docs/stable/sql/statements/analyze
+///
+/// Errors are logged but do not fail the operation since statistics updates are non-critical.
+pub fn execute_analyze_sql(tx: &Transaction, table_name: &str) {
```
sgrebnov: `execute_analyze_sq` is public so it can be re-used, for example for own DuckDB write implementation like table-based partitioned write

lukekim (approved):

PR #481, opened 2025-10-31
Add table provider for ADBC drivers

zeroshade:
What this PR includes:

* `AdbcTableFactory`
* `ADBCPool`
* Some tests (more needed, unit + integration)
* Example (under `examples`)
* CI
* Bulk insert capability

Example:

```rust
let mut driver = ManagedDriver::load_from_name(
        "duckdb", None, AdbcVersion::V110, LOAD_FLAG_DEFAULT, None).unwrap();

let db = driver.
    new_database_with_opts([
          (OptionDatabase::Uri, "core/examples/duckdb_example.db".into())]).unwrap();

let adbc_conn_pool = Arc::new(ADBCPool::new(db, None)
     .expect("unable to create ADBC Conn pool"));

let table_factory = AdbcTableFactory::new(adbc_conn_pool.clone());
let ctx = SessionContext::new();

ctx.register_table(
    "companies_v2",
    table_factory
        .table_provider(TableReference::bare("companies"))
        .await
        .expect("failed to register table provider"),
).expect("failed to register table");

let df = ctx
        .sql("SELECT * FROM datafusion.public.companies_v2")
        .await
        .expect("select failed");
    df.show().await.expect("show failed");
```

As I'm not particularly fluent in Rust yet, I'm sure there's lots of areas for improvement in my rust code and simplifications. I'd love feedback and tips/suggestions from anyone, thanks!

### DIFF
--- a/core/src/sql/db_connection_pool/adbcpool.rs
+++ b/core/src/sql/db_connection_pool/adbcpool.rs
@@ -220,7 +220,7 @@ mod test {
         .unwrap();
 
         driver
-            .new_database_with_opts([(OptionDatabase::Other("path"), ":memory:".into())])
+            .new_database_with_opts([(OptionDatabase::Other("path".to_string()), ":memory:".into())])
             .unwrap()
     }
 


core/src/sql/db_connection_pool/adbcpool.rs
```diff
@@ -0,0 +1,249 @@
...
+    {
+        let pool = Arc::clone(&self.pool);
+        let conn: r2d2::PooledConnection<AdbcConnectionManager<D>> =
+            pool.get().context(ConnectionPoolSnafu)?;
+
+        Ok(Box::new(AdbcDbConnection::new(conn)))
+    }
+
+    fn join_push_down(&self) -> JoinPushDown {
+        // TODO: make this configurable
```
phillipleblanc: This could probably be set to a hash of the connection information, but it can come in a follow up.

phillipleblanc (approved):
This is great @zeroshade! Thanks for this. I had a minor comment about the join push down, but that can come as a follow up.

PR #491, opened 2025-11-17
Readd federation and recursively check children of `Expr` for `FunctionSupport`

Jeadie:


core/src/util/supported_functions.rs
```diff
@@ -107,18 +107,28 @@ impl FunctionSupport {
     }
 
     pub fn supports(&self, expr: &Expr) -> bool {
-        match expr {
-            Expr::ScalarFunction(ScalarFunction { func, .. }) => self.supports_scalar(func),
-            Expr::AggregateFunction(AggregateFunction { func, .. }) => {
-                self.supports_aggregate(func)
+        let mut supports = true;
+        let _ = expr.apply_children(|e| {
```
phillipleblanc: Why use `apply_children` and not `apply`?

PR #498, opened 2025-11-25
Skip uniqueness constraint validation on Overwrite for DuckDB & SQLite

lukekim:
## Summary

Skip uniqueness-constraint validation when performing `InsertOp::Overwrite` for the DuckDB and SQLite table providers. By definition, an overwrite replaces the target table's contents, so pre-existing row values cannot produce duplicate-key violations, and values within the incoming batch are resolved by the write path itself.

## Motivation

The existing write path always called `validate_batch_with_constraints` before inserting, which would reject perfectly valid overwrite batches that happened to contain values conflicting with the *previous* contents of the table (or with themselves, when the caller expected "last write wins" semantics). This made `InsertOp::Overwrite` unusable for common ETL patterns that re-materialize a PK-constrained table from a source.

## Changes

**DuckDB (`core/src/duckdb/`)**
- `write.rs`: Skip `validate_batch_with_constraints` when `self.overwrite == InsertOp::Overwrite`.
- `creator.rs`: Added `create_table_without_constraints`, used for the staging table during overwrite so intra-batch PK duplicates do not trip DuckDB's own constraint check before the atomic swap.

**SQLite (`core/src/sqlite/`, `core/src/sqlite.rs`)**
- `write.rs`: Skip `validate_batch_with_constraints` on `InsertOp::Overwrite`.
- `sqlite.rs`: Thread `InsertOp` through `insert_batch` / `insert_batch_prepared`; Overwrite now uses `REPLACE INTO` so that intra-batch duplicates collapse to the last row written.

**SQL generation (`core/src/sql/arrow_sql_gen/statement.rs`)**
- `InsertBuilder::build` now takes a `replace: bool` flag.
- New `build_sqlite_replace()` helper emits `REPLACE INTO ...` for SQLite overwrite paths.

**Utility (`core/src/util/constraints.rs`)**
- Added `filter_unique_constraints` helper for filtering non-unique constraints out of a `Constraints` set.

## Test coverage

Added four new unit tests in `core/src/duckdb/write.rs` and `core/src/sqlite/write.rs`:

- `test_overwrite_skips_pk_constraint_validation_with_duplicate_pks` (DuckDB & SQLite): writes a batch containing duplicate primary-key values via `InsertOp::Overwrite` and asserts success.
- `test_append_still_enforces_pk_constraint_validation` (DuckDB & SQLite): regression guard — the same duplicate-PK batch must still be rejected when the op is `Append`.

All four pass locally:

```
test sqlite::write::tests::test_overwrite_skips_pk_constraint_validation_with_duplicate_pks ... ok
test duckdb::write::test::test_overwrite_skips_pk_constraint_validation_with_duplicate_pks ... ok
test sqlite::write::tests::test_append_still_enforces_pk_constraint_validation ... ok
test duckdb::write::test::test_append_still_enforces_pk_constraint_validation ... ok
```


core/src/duckdb/write.rs
```diff
@@ -325,15 +325,19 @@ impl DataSink for DuckDBDataSink {
...
-                    constraints,
-                    &crate::util::constraints::UpsertOptions::default(),
-                )
-                .await
-                .context(super::ConstraintViolationSnafu)
-                .map_err(to_datafusion_error)?;
+            // Skip constraint validation for Overwrite operations since we're replacing all data
+            // and uniqueness constraints don't apply to the incoming data in isolation.
+            if self.overwrite != InsertOp::Overwrite {
+                if let Some(constraints) = self.table_definition.constraints() {
```
phillipleblanc, line 331: constraints should still be applied on the incoming data if the final table will have constraints, otherwise you could end up with a table that has violated its constraints.

phillipleblanc: the last write wins should still allow the validation to proceed if that is configured.

PR #504, opened 2025-11-28
Properly handle decoding decimals from SQLite using the decimal extension

phillipleblanc:
Previously we were decoding Decimals from SQLite as UTF8 - this was correct because in the `spiceai` project we were injecting a schema cast step before the datafusion execution kicked in. However, that should logically live in this project so that we can validate the datafusion logic based on the decimal types directly.

### DIFF
--- a/core/src/sql/arrow_sql_gen/sqlite.rs
+++ b/core/src/sql/arrow_sql_gen/sqlite.rs
@@ -392,18 +392,24 @@ mod tests {
 
     #[test]
     fn decode_decimal128_mixed_storage_classes() {
-        let conn = rusqlite::Connection::open_in_memory().unwrap();
-        conn.execute("CREATE TABLE decs (d TEXT)", []).unwrap();
+        let conn = rusqlite::Connection::open_in_memory()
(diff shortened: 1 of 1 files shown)


core/tests/sqlite/mod.rs
```diff
@@ -367,3 +368,153 @@ async fn test_sqlite_table_provider_roundtrip(
         "Round-tripped data should match original"
     );
 }
+
+#[tokio::test]
+async fn test_sqlite_decimal_memory() {
+    let arrow_record =
+        download_parquet_as_record_batch("https://public-data.spiceai.org/decimal.parquet")
```
ewgenius: nitpick: It would be better not to depend on remote data and to include a test data file in the repository

ewgenius (approved):

core/tests/sqlite/mod.rs
phillipleblanc: I view this as equivalent to how other projects do this (i.e. the datafusion-testing repo pulled in via submodule) - you don't really want to commit large files into git, so you need to pull it from somewhere.

PR #521, opened 2026-01-11
feat: Add Oracle table provider with rust-oracle driver

banickn:
Implement Oracle support using rust-oracle (ODPI-C) and bb8 connection pooling. Includes comprehensive type mappings (NUMBER, DATE, TIMESTAMP, CLOB, BLOB, RAW), schema inference, and 12 integration tests.
Tested against Oracle Database 23c Free.

Note: This is my first contribution to the project. Feedback welcome! :)


phillipleblanc:
@banickn thanks again for this — I've pushed some follow-up work directly onto your branch (your original commits are untouched and remain the base of the PR):

- **Rebased onto current trunk.** Main moved providers into per-backend leaf crates after you branched, so the provider now lives in `crates/oracle` as `datafusion-table-providers-oracle`. The bundled `instantclient-basiclite-linuxx64.zip` was dropped (OCI-licensed binaries can't be committed) — CI downloads the Instant Client at runtime instead.
- **Semantics aligned with Spice.ai's production Oracle provider**, which you'd already started borrowing from: Oracle `DATE` → `Timestamp(Second)` (Date32 silently truncated time-of-day; spiceai#12096), unbounded `NUMBER` → `Decimal128(38, 20)`, `TIMESTAMP(0)` → seconds / other precisions → nanoseconds, unsupported types skipped with a warning instead of coerced to `Utf8`, schema lookups scoped via `SYS_CONTEXT('USERENV','CURRENT_SCHEMA')`, default service `XEPDB1`, and exact-integer `BigDecimal` rescaling.
- **Integration tests re-enabled** against the new layout (they needed the `oracle` dev-dependency to compile). They start `gvenzl/oracle-free` in a dedicated CI job and skip cleanly when no database is available. Type expectations were updated to the corrected mappings, and `ALL_TABLES`-based tests now use user-owned fixture tables since schema resolution is owner-scoped.
- Added the crate to the leaf-build matrix and the release/publish workflow.

All pushed commits carry `Co-Authored-By: Norman Banick` so the contribution attribution stays with you. CI is running now.

phillipleblanc:
PR is green on head `9baf731` — all 17 checks pass, including the new **Oracle Integration Tests** job (16/16 tests against gvenzl/oracle-free).

Follow-up commits since the last note:
- `fix(oracle)`: Decimal/Timestamp Arrow builders now carry the field's declared precision/scale and timezone (fixes `column types must match schema types` against a live DB)
- `ci(oracle)`: container readiness via the image's `DATABASE IS READY TO USE!` marker + TCP probe (the image ships no Docker HEALTHCHECK)
- `test(oracle)`: correct N1 expectation (123.456 at scale 20)

@banickn your two original commits are untouched as the base; everything above is co-authored follow-up.

phillipleblanc (approved):

phillipleblanc:
@banickn Thanks!

PR #525, opened 2026-01-13
update datafusion to 52 & update deps

hozan23:


hozan23:
Since dataFusion [removed](https://github.com/apache/datafusion/pull/18528) the `pyarrow` feature, we should migrate to `datafusion-python`. However, version 52 of `datafusion-python` [has not yet been published](https://github.com/apache/datafusion-python/pull/1337), which is why we encountered the previous issue with CI. Let's wait for it.

PR #548, opened 2026-02-14
Add PasswordProvider trait for dynamic Postgres credentials

alexanderbianchi:
## Summary

`PostgresConnectionPool` currently accepts a static password at construction time. The `bb8` pool bakes this credential into every connection it creates. This doesn't work for environments where credentials rotate — which is the norm in production cloud deployments:

- **AWS RDS IAM authentication** — tokens generated via `generate-db-auth-token` expire after **15 minutes**
- **GCP Cloud SQL IAM** — OAuth2 access tokens used as passwords, short-lived
- **Azure AD / Entra** — AAD tokens for Azure Database for PostgreSQL, expire in ~1 hour
- **HashiCorp Vault** — dynamic database credentials with configurable TTLs
- **Kerberos / GSSAPI** — ticket renewal

Today the only workaround is to recreate the entire connection pool when credentials expire, which drops in-flight connections and causes latency spikes. Every mature connection pool has an escape hatch for this — Java's HikariCP has `DataSource`, Go's `database/sql` has `Connector`, Rust's `sqlx` has per-connection `ConnectOptions`. This PR adds the equivalent for `datafusion-table-providers`.

## Changes

Introduces a `PasswordProvider` trait that the pool calls to get a fresh credential each time it creates a new connection:

```rust
#[async_trait]
pub trait PasswordProvider: Send + Sync {
    async fn get_password(&self) -> Result<SecretString>;
}
```

**Example — AWS RDS IAM provider (sketch):**
```rust
struct RdsIamProvider { client: aws_sdk_rds::Client, hostname: String, port: u16, user: String }

#[async_trait]
impl PasswordProvider for RdsIamProvider {
    async fn get_password(&self) -> Result<SecretString> {
        let token = self.client
            .generate_db_auth_token(/* hostname, port, user */)
            .await?;
        Ok(SecretString::from(token))
    }
}
```

### Integration points

- **`PostgresConnectionPool::new_with_password_provider(params, provider)`** — new constructor that accepts an `Arc<dyn PasswordProvider>`
- **`PostgresConnectionPool::new(params)`** — unchanged API; internally wraps the static `pass` param in a `StaticPasswordProvider`, exercising the same code path
- **`ConnectionManager` (custom `bb8::ManageConnection`)** — replaces `bb8_postgres::PostgresConnectionManager`; calls the provider on each `connect()`, only cloning the `Config` when a provider is present (zero overhead for passwordless auth)
- **Credential hygiene** — `get_password()` returns `SecretString` (zeroized on drop); the password is never included in the connection string

### Backward compatibility

- `PostgresConnectionPool::new()` signature and behavior unchanged
- All existing error variants preserved with identical source types
- No new dependencies

## What this does NOT do (future work)

- **Other backends** — `PasswordProvider` is defined in the backend-agnostic `db_connection_pool` module, but this PR only wires it into Postgres. MySQL has the same IAM token pattern on RDS and could reuse the trait in a follow-up.
- **Username rotation** — some IAM schemes (e.g., AWS RDS IAM where the "username" is an IAM role ARN) may also need dynamic usernames. A future `CredentialProvider` returning `(username, password)` could supersede `PasswordProvider`, but the password-only version covers the vast majority of real-world use cases today.
- **IO runtime segregation** — spawning the connection background task on a dedicated IO runtime (for query engines that separate CPU/IO work). Kept out of this PR to limit scope.

## Test plan

- [x] Unit tests: `StaticPasswordProvider` round-trip, `ConnectionManagerError` Display/Error impls, `parse_connection_string` password extraction
- [x] Integration test: creates a pool via `new_with_password_provider` against a real Postgres (Docker), verifies end-to-end through both raw connection and DataFusion `SqlTable` path
- [x] Existing tests pass (roundtrip, one-way, schema inference)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/sql/db_connection_pool/mod.rs
+++ b/core/src/sql/db_connection_pool/mod.rs
@@ -1,6 +1,6 @@
 use async_trait::async_trait;
 use dbconnection::DbConnection;
-use secrecy::{ExposeSecret, SecretString};
+use secrecy::SecretString;
 use std::sync::Arc;
 
 pub mod dbconnection;
@@ -55,7 +55,7 @@ impl StaticPasswordProvider {
 #[async_trait]
 impl PasswordProvider for StaticPasswordProvider {
     async fn get_password(&self) -> Result<SecretString> {
-        Ok(SecretString::from(self.0.expose_secret().to_string()))
+        Ok(self.0.clone())
     }
 }
 


phillipleblanc (approved):

phillipleblanc:
Thanks! This makes sense to me. I pushed up a couple of commits and will merge this.

PR #636, opened 2026-04-28
fix: handle empty projection in SqlTable for count(*) queries

hfldqwe:
## Summary

Fixes a schema mismatch error when executing ` SELECT count(*) ` or ` SELECT count(1) ` queries against SQL-backed tables (MySQL, PostgreSQL, SQLite, etc.).

## Problem

When DataFusion's ` OptimizeProjections ` rule optimizes ` SELECT count(*) ` queries, it reduces the TableScan projection to 0 columns since no columns are referenced. Previously, ` SqlTable ` would return a 1-column schema (` ONE_COLUMN_SCHEMA `) for empty projections (since the SQL ` SELECT 1 ` returns 1 column), causing a physical vs logical schema mismatch:

`
Physical input schema should be the same as the one converted from logical input schema. 
Differences: Different number of fields: (physical) 1 vs (logical) 0
`

This affected ` count(*) `, ` count(1) `, and any aggregation that doesn't reference table columns. ` count(col) ` worked fine because the projection retained at least 1 column.

## Fix

1. **` project_schema_safe() `**: Returns an empty schema (0 columns) for empty projections instead of ` ONE_COLUMN_SCHEMA `. This matches the logical schema that DataFusion expects.

2. **` SqlExec `**: Adds a ` sql_schema ` field that tracks the actual SQL result schema. When projection is empty, ` sql_schema ` is ` ONE_COLUMN_SCHEMA ` (for ` SELECT 1 `) while ` projected_schema ` is empty.

3. **` SqlExec::execute() `**: When projection is empty, executes the SQL with ` sql_schema ` (1 column) and strips the result to 0-column batches that preserve row count, satisfying DataFusion's schema consistency check.

## Testing

Verified with a standalone reproduction against MySQL 8.0:

`
Test 1: SELECT * FROM mysql.testdb.orders LIMIT 1         ... OK (1 rows)
Test 2: SELECT count(id) AS total FROM mysql.testdb.orders ... OK (1 rows)
Test 3: SELECT count(*) AS total FROM mysql.testdb.orders  ... OK (1 rows)  <- was FAILED
Test 4: SELECT count(1) AS total FROM mysql.testdb.orders  ... OK (1 rows)  <- was FAILED
`

All ` SqlExec ` struct construction sites (sort pushdown, limit pushdown, filter pushdown) are updated to propagate the ` sql_schema ` field.

### DIFF
--- a/core/src/sql/sql_provider_datafusion/mod.rs
+++ b/core/src/sql/sql_provider_datafusion/mod.rs
@@ -266,10 +266,12 @@ pub fn project_schema_safe(
     let schema = match projection {
         Some(columns) => {
             if columns.is_empty() {
-                // If the projection is Some([]) then it gets unparsed as `SELECT 1`, so return a schema with a single Int64 column.
-                //
-                // See: <https://github.com/apache/datafusion/blob/83ce79c39412a4f150167d00e40ea05948c4870f/datafusion/sql/src/unparser/plan.rs#L998>
-                Arc::clone(&ONE_COLUMN_SCHEMA)
+                // Empty projection: return empty schema (0 columns).
+                // The SQL will still be `SELECT 1` (SQL syntax requires at least one column),
+                // but the physical output schema must match the logical schema (0 columns)
+                // to satisfy DataFusion's schema consistency check.
+                // The execute() method handles stripping the extra column from the result.
+                Arc::new(Schema::empty())
             } else {
                 Arc::new(schema.project(columns)?)
             }
@@ -281,6 +283,10 @@ pub fn project_schema_safe(
 
 pub struct SqlExec<T, P> {
     projected_schema: SchemaRef,
+    /// Schema of the actual SQL result. When projection is empty, SQL returns
+    /// `SELECT 1` (1 column) but projected_schema is empty (0 columns).
+    /// This field is used to parse the SQL result correctly.
+    sql_schema: SchemaRef,
     pool: Arc<dyn DbConnectionPool<T, P> + Send + Sync>,
     sql: String,
     properties: Arc<PlanProperties>,
@@ -291,6 +297,7 @@ impl<T, P> Clone for SqlExec<T, P> {
     fn clone(&self) -> Self {
         Self {
             projected_schema: Arc::clone(&self.projected_schema),
+            sql_schema: Arc::clone(&self.sql_schema),
             pool: Arc::clone(&self.pool),
             sql: self.sql.clone(),
             properties: self.properties.clone(),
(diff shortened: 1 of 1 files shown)


hfldqwe:
## Bug fix: preserve row count in empty projection batches

The original fix in 8d522e7 used RecordBatch::new_empty(empty_schema) to strip columns from the SQL result, but this was incorrect:

**Problem**: RecordBatch::new_empty() hardcodes 
ow_count to 0. This breaks count(*) queries because DataFusion's aggregation pipeline evaluates COUNT_STAR_EXPANSION (literal Int64(1)) by calling into_array(batch.num_rows()). With 
um_rows() == 0, the literal expands to a 0-length array, and CountAccumulator counts 0 rows instead of the actual count.

**Fix** (8626adb): Use RecordBatch::try_new_with_options() with RecordBatchOptions::with_row_count(Some(b.num_rows())) to create 0-column batches that preserve the original row count from the SQL result.

This ensures that SELECT count(*) FROM table correctly returns the row count even when the projection is empty (0 columns).

Verified by:
- Checking RecordBatch::new_empty source in arrow-rs (row_count is hardcoded to 0)
- Checking DataFusion's CountAccumulator::update_batch and Literal::evaluate + ColumnarValue::into_array(batch.num_rows()) execution chain
- Confirming RecordBatchOptions API is already used elsewhere in this codebase (mysql.rs, postgres.rs, sqlite.rs)

core/src/sql/sql_provider_datafusion/mod.rs
```diff
@@ -266,10 +266,12 @@ pub fn project_schema_safe(
     let schema = match projection {
         Some(columns) => {
             if columns.is_empty() {
-                // If the projection is Some([]) then it gets unparsed as `SELECT 1`, so return a schema with a single Int64 column.
-                //
-                // See: <https://github.com/apache/datafusion/blob/83ce79c39412a4f150167d00e40ea05948c4870f/datafusion/sql/src/unparser/plan.rs#L998>
-                Arc::clone(&ONE_COLUMN_SCHEMA)
+                // Empty projection: return empty schema (0 columns).
```
phillipleblanc, line 269: This was intentionally added to fix the usage of this table provider in the federation use-case, which is its main purpose: https://github.com/datafusion-contrib/datafusion-table-providers/pull/315

phillipleblanc: This change reverts the fix without also fixing it for federation, which we need.

phillipleblanc (changes requested):

hfldqwe:
Closing this PR. After careful analysis, the current behavior on main is correct — PR #315's ONE_COLUMN_SCHEMA fix already handles COUNT(*) properly for both the SqlExec path and the federation fallback path. This PR reverts that fix without need and introduces unnecessary complexity. Thanks for the review @phillipleblanc.

PR #650, opened 2026-05-20
fix(duckdb): use actual DuckDB schema for read provider

ewgenius:
## Problem

`DuckDBTableProviderFactory::create()` uses `cmd.schema` (user-facing types) as the read provider schema. DuckDB silently downgrades certain types during table creation (e.g. `Timestamp(ns, tz)` → `Timestamp(µs, tz)` for `TIMESTAMPTZ`). The read provider then advertises the wrong schema while returning batches with actual DuckDB types, causing:

```
RowConverter column schema mismatch, expected Timestamp(Nanosecond, Some("UTC")) got Timestamp(Microsecond, Some("UTC"))
```

This crashes `ORDER BY` queries on partitioned DuckDB-accelerated datasets.

## Fix

After table creation, query DuckDB for the actual storage schema via `get_schema()` and use that for the read provider. This matches the existing pattern in `DuckDBTableFactory::table_provider()`.

## Test

Added `test_read_provider_schema_reflects_actual_duckdb_types` which verifies the read provider advertises `Timestamp(Microsecond, ...)` when DuckDB downgrades from the requested `Timestamp(Nanosecond, ...)`.

### DIFF
--- a/core/src/duckdb.rs
+++ b/core/src/duckdb.rs
@@ -1178,9 +1178,7 @@ pub(crate) mod tests {
         // DuckDB stores TIMESTAMPTZ as Microsecond regardless of requested precision.
         match ts_field.data_type() {
             DataType::Timestamp(TimeUnit::Microsecond, _) => {}
-            other => panic!(
-                "Expected Timestamp(Microsecond, _), got {other:?}"
-            ),
+            other => panic!("Expected Timestamp(Microsecond, _), got {other:?}"),
         }
     }
 }
--- a/core/src/duckdb.rs
+++ b/core/src/duckdb.rs
@@ -485,15 +485,7 @@ impl TableProviderFactory for DuckDBTableProviderFactory {
                 .with_indexes(indexes.clone());
 
         let pool = Arc::new(pool);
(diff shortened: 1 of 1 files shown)


core/src/duckdb.rs
phillipleblanc, line 480: The fix should read the duckdb schema into this `schema` variable, which is also used in `TableDefinition::new`

ewgenius: Thanks, updated to have `schema` reassigned to actual schema after table is created, and reused in all places.

core/src/duckdb.rs
```diff
@@ -485,15 +485,7 @@ impl TableProviderFactory for DuckDBTableProviderFactory {
                 .with_indexes(indexes.clone());
```
phillipleblanc, line 485: Isn't this dead code now?

ewgenius: No, it is used to create initial table

https://github.com/datafusion-contrib/datafusion-table-providers/blob/38c47b6b9382584b37d33453269bebddc6231fb3/core/src/duckdb.rs#L482-L489

phillipleblanc (approved):

PR #653, opened 2026-05-21
fix(DuckDB): replace `cast_batch_to_schema` with federation's `try_cast_to`

sgrebnov:
Replace `cast_batch_to_schema` with `datafusion_federation::schema_cast::record_convert::try_cast_to`, which is more robust and supports handling interval cross-type casts, string->list, string->struct, and other edge cases.


core/tests/duckdb/mod.rs
```diff
@@ -68,7 +67,6 @@ async fn arrow_duckdb_round_trip(
         .expect("DataFrame should be created from query");
 
     let record_batch = df.collect().await.expect("RecordBatch should be collected");
-    let casted_record = try_cast_to(record_batch[0].clone(), source_schema).unwrap();
```
sgrebnov, line 71: This is not required anymore as duckdb fully respects the schema (uses the same logic internally)

ewgenius (approved):

PR #676, opened 2026-07-03
Upgrade sea-query to 1.0

nuno-faria:
`sea-query` 1.0 exposes new features which are useful downstream. Also made `table_reference_to_sea_table_ref` public so it can be reused by `sqlite.rs`.

### DIFF
--- a/core/src/sql/arrow_sql_gen/statement.rs
+++ b/core/src/sql/arrow_sql_gen/statement.rs
@@ -14,8 +14,9 @@ use num_bigint::BigInt;
 use sea_query::{
     Alias, ColumnDef, ColumnType, Expr, GenericBuilder, Index, InsertStatement, IntoIden,
     IntoIndexColumn, Keyword, MysqlQueryBuilder, OnConflict, PostgresQueryBuilder, Query,
-    QueryBuilder, SeaRc, SimpleExpr, SqliteQueryBuilder, Table, TableRef,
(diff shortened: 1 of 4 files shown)


core/src/sql/arrow_sql_gen/statement.rs
```diff
@@ -1116,24 +1117,17 @@ impl<'a> InsertBuilder<'a> {
...
             schema,
             table,
-        } => TableRef::DatabaseSchemaTable(
-            SeaRc::new(Alias::new(catalog.to_string())),
-            SeaRc::new(Alias::new(schema.to_string())),
-            SeaRc::new(Alias::new(table.to_string())),
-        ),
+        } => (catalog.to_string(), schema.to_string(), table.to_string()).into_table_ref(),
     }
 }
```
nuno-faria, line 1132: `TableRef` has been changed, so the previous code does not work. Instead, we can now simply use `into_table_ref`. See: https://github.com/SeaQL/sea-query/pull/927.

nuno-faria:
@phillipleblanc would it be possible to cut a release with this before upgrading to DataFusion 54?

phillipleblanc (approved):

phillipleblanc:
@nuno-faria I will cut a new release with this later today before merging the DF54 upgrade PR

nuno-faria:
Thank you!

PR #680, opened 2026-07-05
Compile integration tests in CI

nuno-faria:
I noticed there were errors when trying to compile and run the integration tests, so I think it's better to at least compile them in CI to avoid future errors.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -112,7 +112,7 @@ r2d2_adbc = { version = "0.3.0", optional = true }
 
 [dev-dependencies]
 anyhow = "1.0"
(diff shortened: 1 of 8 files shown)


core/Cargo.toml
```diff
@@ -112,7 +112,7 @@ r2d2_adbc = { version = "0.3.0", optional = true }
 
 [dev-dependencies]
 anyhow = "1.0"
-bollard = "0.20"
+bollard = "=0.19.4"
```
nuno-faria, line 115: Fixed to `0.19.4` as the API has since changed.

phillipleblanc (approved):

PR #686, opened 2026-07-18
feat(postgres): thread libpq `options` connection param onto the Config

greg-semvia:
## What

`PostgresConnectionPool::new` already threads a `application_name` connection
param onto the `tokio_postgres::Config`. This adds the same treatment for the
libpq `options` param, so a caller can pass e.g.
`options = "-c statement_timeout=5000"` and have it reach the server as a
per-session `SET`.

Both params are now applied by one small helper,
`apply_optional_session_params`, keeping `new_inner` tidy:

```rust
fn apply_optional_session_params(config: &mut Config, params: &HashMap<String, SecretString>) {
    if let Some(application_name) = params.get("application_name").map(SecretBox::expose_secret) {
        config.application_name(application_name);
    }
    if let Some(options) = params.get("options").map(SecretBox::expose_secret) {
        config.options(options);
    }
}
```

## Why

`options` is the standard libpq escape hatch for per-connection server
settings (`statement_timeout`, `lock_timeout`, `search_path`, …). Surfacing it
lets a governance layer bound query execution time at the source without a
custom pool wrapper. `application_name` was already special-cased; this widens
that path by exactly one well-understood param rather than adding a bespoke
knob.

## Tests

Two unit tests over the pure config assembly (no live server needed):

- `options_param_reaches_the_config` — asserts `config.get_options()` and
  `config.get_application_name()` reflect the params.
- `absent_options_param_leaves_the_config_untouched` — asserts the default
  (`None`) is preserved when the param is absent.

Run with:

```bash
cargo test --no-default-features --features postgres-federation --lib postgrespool
```

### DIFF
--- a/core/src/sql/db_connection_pool/postgrespool.rs
+++ b/core/src/sql/db_connection_pool/postgrespool.rs
@@ -351,10 +351,7 @@ impl PostgresConnectionPool {
         let mut config =
             Config::from_str(connection_string.as_str()).context(ConnectionPoolSnafu)?;
 
-        if let Some(application_name) = params.get("application_name").map(SecretBox::expose_secret)
-        {
-            config.application_name(application_name);
-        }
+        apply_optional_session_params(&mut config, &params);
 
         verify_postgres_config(&config).await?;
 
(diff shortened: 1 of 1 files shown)


phillipleblanc (approved):
This makes sense to me, thanks!

PR #720, opened 2026-08-14
Fix silent precision loss on unconstrained Postgres numeric columns

### LINKED ISSUE #719: Postgres: unconstrained `numeric` column is rounded to the first row's scale (silent precision loss)
## Describe the bug

For an **unconstrained** Postgres `numeric` column, the Arrow decimal scale is taken from the first non-null row of the result set and every subsequent value is `rescale`d down to it. `rescale` rounds, so values with more decimal places than the first row are silently corrupted.

Columns with a **declared** scale (`numeric(10,3)`) are unaffected — they take the schema-driven path in `get_decimal_column_precision_and_scale`. The problem is specific to unconstrained `numeric`, which is what `max()`, `min()`, `avg()` and arithmetic over a `numeric` column return, so aggregates are the common exposure.

## To Reproduce

```sql
SELECT v FROM (VALUES (1.0::numeric),(2.5::numeric),(3.75::numeric)) t(v)
-- returns 1.0, 2.5, 3.8        ← 3.75 rounded to the first row's 1dp

SELECT v FROM (VALUES (3.75::numeric),(1.0::numeric),(2.5::numeric)) t(v)
-- returns 3.75, 1.0, 2.5       ← same values, reordered, all exact
```

Worst case, an integer first row gives the column zero decimal places:

```sql
SELECT v FROM (VALUES (20::numeric),(17.685::numeric),(15.334::numeric)) t(v)
-- returns 20.0, 18.0, 15.0     ← two values destroyed
```

A single-row result is always exact, so this survives casual testing.

## Expected behavior

Every value returned exactly, independent of row order.

## Source

`src/sql/arrow_sql_gen/postgres.rs`, the `Type::NUMERIC` arm (0.11.0). The existing comment identifies the cause:

```rust
// Record Batch Scale is determined by first row, while Postgres Numeric Type doesn't have fixed scale
// Resolve scale difference for incoming records
let dest_scale = postgres_numeric_scale.unwrap_or_default();
v.rescale(dest_scale);            // lossy when dest_scale < v.scale()
dec_builder.append_value(v.mantissa());
```

`postgres_numeric_scale` is set once from the first row (`if postgres_numeric_scale.is_none()`), and `Decimal::rescale` rounds when narrowing.

## Suggested fix

The scale must never narrow below what a value needs. Options:

1. **Two-pass per batch** — scan the batch for `max(scale)` and build the `Decimal128Builder` with that. Correct within a batch; still wrong across batches unless the field is fixed per stream.
2. **Pin unconstrained `numeric` to a fixed generous scale** (e.g. the `Decimal128(38, N)` the builder already uses for precision), so `rescale` only ever widens. Simple and order-independent; costs nothing for typical values.
3. **Error rather than round** if a value cannot be represented at the batch's scale, so the loss is at least visible.

Option 2 seems the smallest correct change. Whichever is chosen, a regression test needs the **first row to be a whole number** and later rows to carry more decimals — the bug is invisible if the fixture is ordered the other way.

## Impact seen in practice

Found in a BMS deployment reading a 91 GB Timescale table whose `value` is `numeric`. A sensor frozen at `15.33` was reported as `15.0`, and `17.68` as `18.0`; a real temperature series `20.0, 19.82, 19.64, …` came back as `20.0, 19.0, 19.0, …`. Alerting logic that compares a transported value against a threshold makes wrong decisions, not merely wrong display.

Workaround for callers: cast `::float8` or `::text` in the projection.

NubeDev:
## Which issue does this PR close?

Closes #719.

## Rationale for this change

An unconstrained Postgres `numeric` column has no fixed scale — each row may carry a different number of decimal places — but the Arrow column needs a single scale. `rows_to_arrow` took that scale from the first non-null row and then `Decimal::rescale`d every later value down to it. `rescale` rounds, so longer values were silently corrupted:

```sql
SELECT v FROM (VALUES (20::numeric),(17.685::numeric),(15.334::numeric)) t(v)
-- returned 20, 18, 15
```

The same values in a different order came back exact, and a single-row result is always exact, so the loss survives casual testing. `max()`, `min()`, `avg()` and arithmetic over a `numeric` column all return this unconstrained type, so aggregates were the common exposure. Columns with a declared scale (`numeric(10,3)`) were unaffected — they take the schema-driven path.

## What changes are included in this PR?

- The Arrow scale for a `numeric` column that declares none is pinned to `Decimal128(38, 20)` instead of being inferred from the first row. That is the same scale schema inference already assigns to a bare `numeric`, so the row path and the schema path now agree, and the result no longer depends on row order or on how rows are chunked into batches.
- Values are converted to the column's mantissa by exact `i128` arithmetic (`decimal_to_i128_mantissa`) rather than `Decimal::rescale`, which rounds when narrowing and silently refuses to widen past its own 96-bit mantissa. A value that cannot be represented exactly at the column scale is now an error (`DecimalExceedsColumnScale` / `DecimalOverflowsColumnScale`) instead of a quietly rounded value.
- `numeric[]` derived its scale from the first row's widest element and had the same order dependence; it is fixed the same way.
- Columns with a declared scale keep taking their scale from the projected schema, unchanged.

## Are there any user-facing changes?

Yes, and they are intended:

- Unconstrained `numeric` columns now always come back as `Decimal128(38, 20)` rather than at whichever scale the first row happened to have. Values that were previously rounded are now exact.
- A `numeric` value with more decimal places than the column's scale is rejected with an error instead of being silently rounded. Previously this loss was invisible.

No public API changes.

## Testing

- `core/tests/postgres/mod.rs`: `test_postgres_unconstrained_numeric_precision` runs the issue's repro through `query_arrow` with no projected schema and asserts both the column scale and every value. The fixture is deliberately ordered whole-number first (`20, 17.685, 15.334`) — ordered the other way it passes against the unfixed code. Verified failing before the fix (`left: 0, right: 20`, i.e. the zero-decimal worst case) and passing after.
- `crates/postgres/src/arrow_sql_gen/mod.rs`: unit tests for `decimal_to_i128_mantissa` covering exact widening, an already-matching declared scale, refusal to round, and 128-bit overflow.
- `cargo fmt --all -- --check`, `cargo clippy -p datafusion-table-providers-postgres --all-features -- -D warnings`, `cargo test -p datafusion-table-providers-postgres --lib` (57 passed), and the full Docker-backed Postgres integration suite `cargo test -p datafusion-table-providers --features postgres --test integration postgres` (21 passed) all clean locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/postgres/src/arrow_sql_gen/mod.rs
+++ b/crates/postgres/src/arrow_sql_gen/mod.rs
@@ -101,14 +101,23 @@ pub enum Error {
     },
 
     #[snafu(display(
-        "Cannot represent the Postgres numeric value {value} as Decimal128(38, {scale}): the value overflows 128 bits"
+        "Cannot represent the Postgres numeric value {value} in column '{column_name}' as Decimal128({precision}, {scale}): the value overflows 128 bits"
     ))]
-    DecimalOverflowsColumnScale { value: String, scale: u32 },
+    DecimalOverflowsColumnScale {
+        column_name: String,
+        value: String,
+        precision: u8,
+        scale: u32,
+    },
 
     #[snafu(display(
-        "The Postgres numeric value {value} has more decimal places than the column scale {scale}; rescaling it would silently lose precision"
+        "The Postgres numeric value {value} in column '{column_name}' has more decimal places than the column scale {scale}; rescaling it would silently lose precision"
     ))]
-    DecimalExceedsColumnScale { value: String, scale: u32 },
+    DecimalExceedsColumnScale {
+        column_name: String,
+        value: String,
+        scale: u32,
+    },
 
     #[snafu(display("The field '{field_name}' has an unsupported data type: {data_type}."))]
     UnsupportedDataType {
@@ -119,17 +128,14 @@ pub enum Error {
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-/// The Arrow scale used for a Postgres `numeric` column that declares no scale of its own
-/// (bare `numeric`, and anything derived from it such as `max()`, `avg()` or arithmetic).
-///
-/// Such a column has no fixed scale in Postgres: every row may carry a different number of
-/// decimal places. The Arrow column, however, needs one scale for all rows, so it is pinned
-/// here instead of being inferred from whichever row happens to arrive first — inferring it
-/// made the result depend on row order and silently rounded every later, longer value.
-///
-/// This matches the scale schema inference already assigns to a bare `numeric`
-/// (`Decimal128(38, 20)`), so the row path and the schema path agree.
-const DEFAULT_UNCONSTRAINED_NUMERIC_SCALE: u32 = 20;
+/// The Arrow precision for a `numeric` column that declares none, taken from schema inference
+/// so the two paths cannot drift apart — see [`schema::DEFAULT_NUMERIC_PRECISION`].
+const DEFAULT_NUMERIC_PRECISION: u8 = schema::DEFAULT_NUMERIC_PRECISION;
+
+/// The scale for such a column, as the unsigned column scale this decode path carries. Same
+/// source of truth as the precision above; [`schema::DEFAULT_NUMERIC_SCALE`] documents why it
+/// is pinned rather than inferred from the first row.
+const DEFAULT_UNCONSTRAINED_NUMERIC_SCALE: u32 = schema::DEFAULT_NUMERIC_SCALE as u32;
 
 /// Returns the `Decimal128` mantissa representing `value` in a column of scale `dest_scale`,
 /// without losing precision.
@@ -138,7 +144,10 @@ const DEFAULT_UNCONSTRAINED_NUMERIC_SCALE: u32 = 20;
 /// silent corruption this replaces) and refuses to widen past its own 96-bit mantissa. Both
 /// directions are handled explicitly here, and anything that cannot be represented exactly is
 /// reported as an error rather than quietly rounded.
-fn decimal_to_i128_mantissa(value: Decimal, dest_scale: u32) -> Result<i128> {
+///
+/// `column_name` names the offending column in that error — a numeric that will not fit is a
+/// property of one column of the query, and a message without it leaves the user to guess which.
+fn decimal_to_i128_mantissa(value: Decimal, dest_scale: u32, column_name: &str) -> Result<i128> {
     let value_scale = value.scale();
     let mantissa = value.mantissa();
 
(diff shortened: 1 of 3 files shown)


crates/postgres/src/arrow_sql_gen/mod.rs
```diff
@@ -100,6 +100,16 @@ pub enum Error {
         source: arrow::error::ArrowError,
     },
 
+    #[snafu(display(
```
bjchambers, line 103: Can we put the column name in these two errors? It would make it easier for a user to understand and address.

crates/postgres/src/arrow_sql_gen/mod.rs
```diff
@@ -109,6 +119,48 @@ pub enum Error {
...
+/// (bare `numeric`, and anything derived from it such as `max()`, `avg()` or arithmetic).
+///
+/// Such a column has no fixed scale in Postgres: every row may carry a different number of
+/// decimal places. The Arrow column, however, needs one scale for all rows, so it is pinned
+/// here instead of being inferred from whichever row happens to arrive first — inferring it
+/// made the result depend on row order and silently rounded every later, longer value.
+///
+/// This matches the scale schema inference already assigns to a bare `numeric`
+/// (`Decimal128(38, 20)`), so the row path and the schema path agree.
+const DEFAULT_UNCONSTRAINED_NUMERIC_SCALE: u32 = 20;
```
bjchambers: The constant should be shared between schema inference and this file to ensure they stay in sync.

bjchambers (commented):
Thanks for the bug report and the fix. A few minor comments / questions.

NubeDev:
Thanks for the review — both addressed in 676ef65.

**Column name in the errors.** `DecimalExceedsColumnScale` and `DecimalOverflowsColumnScale` now carry a `column_name`, threaded from `column_names[i]` at both the `NUMERIC` and `NUMERIC_ARRAY` decode sites, so a user hitting either sees which column of their query is at fault and not just the value:

```
The Postgres numeric value 1.235 in column 'reading' has more decimal places than the column scale 2; rescaling it would silently lose precision
Cannot represent the Postgres numeric value 79228162514264337593543950335 in column 'reading' as Decimal128(38, 20): the value overflows 128 bits
```

The overflow message also reports the precision it could not fit rather than hardcoding 38 in the text. The unit test asserts the column name is present in both.

**Shared constant.** `DEFAULT_NUMERIC_PRECISION` / `DEFAULT_NUMERIC_SCALE` now live next to schema inference in `arrow_sql_gen/schema.rs`, and both `parse_numeric_type` (which previously hardcoded `(38, 20)`) and the row-decode path in `mod.rs` read them. The two paths agreeing is exactly what makes this fix work, so this is worth pinning — thanks for catching it.

Re-ran here: `cargo fmt --all -- --check` clean, `cargo clippy -p datafusion-table-providers-postgres --all-features -- -D warnings` clean, `--lib` 57 passed, and the Docker Postgres integration suite 21 passed (including the new `numeric` regression case inside `test_arrow_postgres_one_way`).

phillipleblanc (approved):

PR #744, opened 2026-09-15
fix(tests): Remove the Docker containers after test execution

martin-g:
* Use `linktime` crate to setup constructor/destructors for global setup/teardown in the IT tests (postgres, mysql and clickhouse). Using rstest's `[once]` didn't work because ConnectionManager's Drop was not called.
* Control whether the Docker containers should be removed or just stopped at the end via  `DF_TABLE_PROVIDERS_DEBUG` env var. Any value will just stop the Docker container. This way the developer could re-start it manually and debug something if needed.
* Use u16 for the `port` variable


martin-g:
Changed to `Draft` to do the same improvements for MongoDB.

core/tests/integration.rs
```diff
@@ -1,16 +1,12 @@
...
 mod adbc;
 mod arrow_record_batch_gen;
 #[cfg(feature = "clickhouse")]
 mod clickhouse;
-#[cfg(any(
-    feature = "clickhouse",
-    feature = "mongodb",
-    feature = "mysql",
-    feature = "postgres"
-))]
```
martin-g: This is removed to because otherwise building any of the other features fail to find `crate::docker::RunningContainer`.
Please let me know if you think this is the wrong approach and I will try to rework it!

PR #745, opened 2026-09-16
chore: Remove unused dependencies

martin-g:
The dependencies are reported as unused by RustRover and/or `cargo-machete`

Add a CI workflow to detect this issue in the future


martin-g:
Note: the new `machete.yaml` CI workflow is not executed because it needs to be in `main` to be active.

